PR #22, opened 2017-07-12
CircleCI 2.0 support

### LINKED ISSUE #20: Ubuntu test environments
We should decide whether the gitlab setting (16.04 LTS + 17.04 latest) should be tested or whether we keep the current setting (16.10).

FloopCZ:
closes #20
supersedes #21


FloopCZ:
closes #20 
supersedes #21

petrbel:
readme should be updated so that it provides the correct information about tested environments

blazekadam:
Please change CI to :latest and :rolling instead of fixed versions.

PR #23, opened 2017-07-13
Continuous deployment

blazekadam:
This PR implements PyPI continuous deployment in CirclesCI.

Every push->build in the master branch launches package upload to PyPI via my account which is set up with ENV variables defined in Circles WebUI.

From now on, we should strictly follow this workflow:
- branch from dev
- implement new features, fix bugs
- create PR to dev

When appropriate (one or more changes merged to dev)
- increase the version in setup.py according to [Semantic Versioning 2.0](http://semver.org/)
- PR -> merge to dev to master
- make a github release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@ from setuptools import setup
 
 setup(name='cxflow',
       version='0.6',
-      description='Smart TensorFlow trainer',
-      long_description='Trainer of TensorFlow models that automatically manages the whole process of training,'
+      description='Smart neural network trainer',
+      long_description='Trainer of neural network models that automatically manages the whole process of training,'
                        'saving and restoring models and much more',
       classifiers=[
         'Development Status :: 4 - Beta',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -19,10 +19,10 @@
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6'
       ],
-      keywords='tensorflow wrapper',
-      url='https://gitlab.com/Cognexa/cxflow',
+      keywords='neural-networks training framework',
+      url='https://github.com/cognexa/cxflow',
       author='Petr Belohlavek',
```
petrbel: we should change the authors in all our projects...

blazekadam: How would you change them?

petrbel: Multiple options are possible, e.g. specifying the full list of contributors. However, I prefer creating an `cxflow` alias to my/your/... mail or creating a mailing list as [this post](http://bugs.python.org/msg93284) suggests.

PR #40, opened 2017-07-26
Update setup.py author credentials

blazekadam:


setup.py
```diff
@@ -21,8 +21,8 @@
       ],
       keywords='neural-networks training framework',
```
FloopCZ: why do we use hyphen in `neural-networks`?

blazekadam: what hyphen?

PR #42, opened 2017-07-29
ProfileHook fix & tests

### LINKED ISSUE #41: profiling hook is inconsistent with mainloop paradigm
The profiling hook requires valid stream to be present although this is not required by the main loop.

This issue is similar to #5

blazekadam:
Changes
- add `extra_streams` to `after_epoch_profile` event so the profile may be handled with ease
- add regression tests for #41
- fix `ProfileHook`
- add logging contents tests

closes #41

### DIFF
--- a/cxflow/main_loop.py
+++ b/cxflow/main_loop.py
@@ -167,7 +167,8 @@ class MainLoop:   # pylint: disable=too-many-instance-attributes
                         hook.after_epoch(epoch_id=epoch_id, epoch_data=epoch_data)
 
                 for hook in self._hooks:
-                    hook.after_epoch_profile(epoch_id=epoch_id, profile=self._epoch_profile)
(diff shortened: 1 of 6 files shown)


cxflow/main_loop.py
```diff
@@ -167,7 +167,8 @@ def run(self) -> None:
                         hook.after_epoch(epoch_id=epoch_id, epoch_data=epoch_data)
 
                 for hook in self._hooks:
-                    hook.after_epoch_profile(epoch_id=epoch_id, profile=self._epoch_profile)
+                    hook.after_epoch_profile(epoch_id=epoch_id, profile=self._epoch_profile,
+                                             extra_streams=self._extra_streams)
```
petrbel: Why not to pass `extra_streams` to the constructor instead?

PR #43, opened 2017-08-01
TrainCheckHookTest done.

gnovis:


cxflow/tests/hooks/train_check_hook_test.py
```diff
@@ -0,0 +1,75 @@
...
+            hook.after_epoch(_CURRENT_EPOCH, self._epoch_data)
+
+    def test_epoch_raise(self):
+        """
+        Test raising error, when required value of
+        stream variable wasn't reached in specified count of epochs.
+        """
+        hook = TrainCheckHook(_VAR, _MIN_ACCURACY, _MAX_EPOCH, "test")
+        with self.assertRaises(ValueError):
+            hook.after_epoch(11, self._epoch_data)
```
blazekadam: No newline at the end of the file :) PEP8

blazekadam: Other than than, it looks good. Thanks!

cxflow/tests/hooks/train_check_hook_test.py
```diff
@@ -0,0 +1,75 @@
...
+    ])
+    return epoch_data
+
+
+class TrainCheckHookTest(CXTestCase):
+    """Test case for TrainCheckHook."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._epoch_data = _get_epoch_data()
```
petrbel: I think that `_get_epoch_data()` should be invoked in each test separately instead of "caching" it in the constructor. That is because hooks (in general) might modify the `epoch_data`. This hook is rather special, however, in order to set a good example I prefer not to cache the data at all.

PR #51, opened 2017-08-05
Fix profile hook

gnovis:


cxflow/hooks/profile_hook.py
```diff
@@ -22,7 +22,7 @@ class ProfileHook(AbstractHook):
     -------------------------------------------------------
     """
 
-    def after_epoch_profile(self, _, profile: Timer.TimeProfile, extra_streams: Iterable[str]) -> None:
+    def after_epoch_profile(self, epoch_id, profile: Timer.TimeProfile, extra_streams: Iterable[str]) -> None:
```
petrbel: I've bumped into it few days ago. Actually don't know why we hadn't noticed before. We use this thing!

PR #54, opened 2017-08-07
Removed create_[valid/test]_stream

gnovis:


blazekadam:
This is misunderstanding. The methods should not be removed. We want to get rid of the `create_` prefix. I.e., its only a refactor. Besides the datasets, `main_loop.py` needs to be changed.

petrbel:
btw mainloop should be updated as well, doesn't it?

blazekadam:
I just pointed that out :)

petrbel:
oh :D sorry

PR #56, opened 2017-08-07
Dataset nesting removed

gnovis:


cxflow/entry_point.py
```diff
@@ -104,7 +104,9 @@ def create_dataset(config: dict, output_dir: Optional[str]=None) -> AbstractData
     :return: dataset object
     """
     logging.info('Creating dataset')
-    dataset_config = {'dataset': config['dataset']}
+    dataset_config = config['dataset']
+    if 'output_dir' in dataset_config:
+        raise ValueError("The 'output_dir' key is reserved and can not be used in dataset configuration.")
```
blazekadam: Use apostrophes and back-ticks respectively.

cxflow/entry_point.py
```diff
@@ -104,7 +104,9 @@ def create_dataset(config: dict, output_dir: Optional[str]=None) -> AbstractData
...
     if output_dir:
```
blazekadam: Remove the condition.

PR #85, opened 2017-08-25
Remove name from model `**kwargs`

blazekadam:


cxflow/cli/common.py
```diff
@@ -102,16 +102,18 @@ def create_model(config: dict, output_dir: str, dataset: AbstractDataset,
...
+    if 'name' in model_kwargs:
+        del model_kwargs['name']
 
     try:
         model = create_object(model_module, model_class, kwargs=model_kwargs)
     except (ImportError, AttributeError) as ex:
         if restore_from is None:  # training case
             raise ImportError('Cannot create model from the specified model module `{}` and class `{}`.'.format(
-                model_config['module'], model_config['class'])) from ex
+                model_config, model_config)) from ex
```
FloopCZ: This does not seem to be right. I suppose that you wanted to replace those with `model_module` and `model_class`.

blazekadam: True, apparently this line is not covered in the tests...

PR #88, opened 2017-08-27
Add get_attribute function

blazekadam:


petrbel:
do we want to tests this?

blazekadam:
It is implicitly tested by the `create_object` test.

PR #90, opened 2017-08-28
Doc gen refactor

Teyras:


docs/conf.py
```diff
@@ -1,51 +1,8 @@
...
-    'sphinx.ext.inheritance_diagram',
-    'sphinx.ext.intersphinx',
-    'sphinx.ext.todo',
-    'sphinx.ext.coverage',
-    'sphinx.ext.ifconfig',
-    'sphinx.ext.mathjax',
-    'sphinx.ext.viewcode',
-    'sphinx_autodoc_typehints']
+sys.path.insert(0, '_base')
+from conf import *
```
FloopCZ: Ugly but necessary, I guess. If so, it is a nice hack. ;-)

PR #91, opened 2017-08-28
move pypi upload to a separate repo

### LINKED ISSUE #60: Use CircleCI deployment
instead of running at the end of the arch tests

Teyras:
Closes #60


.circleci/config.yml
```diff
@@ -89,6 +88,16 @@ jobs:
...
       - *doc
 
+  deploy:
+    docker:
+      - image: pritunl/archlinux:2017-07-08
+    working_directory: ~/cxflow
+    steps:
+      - *arch_deps
+      - checkout
+      - *install
```
FloopCZ: Is it really necessary to install it when building the wheel?

.circleci/config.yml
```diff
@@ -105,3 +114,11 @@ workflows:
...
             - test_ubuntu_rolling
             - test_archlinux
+      - deploy:
+          filters:
+            branches:
+              only: master
+          requires:
+            - test_ubuntu_latest
+            - test_ubuntu_rolling
+            - test_archlinux
```
petrbel: I suggest adding `generate_documentation` as a deploy dependency as we probably want the up2date documentation

petrbel:
@Teyras this PR closes #60, doesn't it?

.circleci/config.yml
FloopCZ: You are right. We don't want to deploy the wheel when sphinx fails.

PR #109, opened 2017-09-03
Progress hook

gnovis:
The hook which displays progress of epoch.

### DIFF
--- a/cxflow/hooks/progress.py
+++ b/cxflow/hooks/progress.py
@@ -1,0 +1,53 @@
+from . import AbstractHook
+from ..datasets import AbstractDataset
+
+_ERASE_LINE = '\x1b[2K'
+
+
+class Progress(AbstractHook):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._batch_count = {}
+        self._first_batch_in_epoch = True
+        self._first_epoch = True
+        self._batch_count_saved = None
+
+    def after_batch(self, stream_name: str, batch_data: AbstractDataset.Batch):
+
+        if stream_name not in self._batch_count:
+            self._batch_count[stream_name] = 0
+        else:
+            self._batch_count[stream_name] += 1
+
+        if not self._first_batch_in_epoch:
+            print(_ERASE_LINE, end='\r')
+        else:
+            self._first_batch_in_epoch = False
+
+        current_batch = self._batch_count[stream_name] + 1
+
+        if self._batch_count_saved:
+            total_batches = self._batch_count_saved[stream_name] + 1
+            self._print_progress_bar(
+                current_batch,
+                total_batches,
+                prefix='Progress of {} stream:'.format(stream_name), suffix='complete')
+        else:
+            print("Progress of {} stream: {} batch complete".format(stream_name,
+                                                                    current_batch), end='\r')
+
+    def _reset(self):
+        self._batch_count = {}
+        self._first_batch = True
+
+    def after_epoch(self, **_):
+        if self._first_epoch:
+            self._batch_count_saved = self._batch_count.copy()
+            self._first_epoch = False
+        self._reset()
+
+    def _print_progress_bar(self, iteration, total, prefix='', suffix='', length=50, fill='█'):
+        percent = ("{0:.2f}").format(100 * (iteration / float(total)))
+        filled_len = int(length * iteration // total)
+        bar = fill * filled_len + '-' * (length - filled_len)
+        print('\r%s |%s| %s/%s %s%% %s' % (prefix, bar, iteration, total, percent, suffix), end='\r')
--- a/cxflow/hooks/progress.py
+++ b/cxflow/hooks/progress.py
@@ -1,41 +1,38 @@
 """
-Module with the hook which displays progress of the current epoch.
+Module with the hook which shows progress of the current epoch.
 """
 
 import shutil
+import collections
 
 from . import AbstractHook
 from ..datasets import AbstractDataset
 
 
-class Progress(AbstractHook):
+class ShowProgress(AbstractHook):
     """
-    Display progress of a processed stream in the current epoch.
+    Show progress of a processed stream in the current epoch.
 
     .. code-block:: yaml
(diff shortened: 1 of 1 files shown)


cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Create new Progress hook.
+        """
+        super().__init__(**kwargs)
+        self._batch_count = {}
+        self._batch_count_saved = None
+        self._first_batch_in_epoch = True
```
petrbel: setting `self._batch_count` and `self._first_batch_in_epoch` is already done in `_reset`: wouldn't it be nicer to omit these lines in the constructor and simply run `_reset` instead?

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+        param fill: char to be displayed as a step in the progress bar
+        """
+
+        percent = ("{0:.1f}").format(100 * (iteration / float(total)))
+        filled_len = int(length * iteration // total)
+        bar = fill * filled_len + '-' * (length - filled_len)
+
+        print('\r%s |%s| %s/%s=%s%% %s' % (prefix, bar, iteration, total, percent, suffix), end='\r')
+
+    def _erase_line(self):
```
petrbel: this should be static method or function entirely out of class (btw pycharm suggests this automatically)

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def after_batch(self, stream_name: str, batch_data: AbstractDataset.Batch):
+        """
+        For the first epoch just the count of processed batches is displayed, because the size of the
+        current stream is unknown. For the following epochs the progress bar is displayed.
+
+        param stream_name: name of current stream
+        """
+
+        if stream_name not in self._batch_count:
```
blazekadam: This can be shortened with defaultdict.

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+
+        param iteration: current step in computation
+        param total: total count of steps in computation
+        param prefix: info text displayed before the progress bar
+        param suffix: info text displayed after the progress bar
+        param length: length of progress bar
+        param fill: char to be displayed as a step in the progress bar
+        """
+
+        percent = ("{0:.1f}").format(100 * (iteration / float(total)))
```
petrbel: redundant parenthesis (btw pycharm...)

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+        self._first_batch_in_epoch = True
+
+    def _reset(self) -> None:
+        """
+        Reset `_batch_count` and `_first_batch_in_epoch` to initial values.
+        """
+        self._batch_count = {}
+        self._first_batch_in_epoch = True
+
+    def _print_progress_bar(self, iteration: int, total: int, prefix: str='',
```
petrbel: this should be also static method

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+        Create new Progress hook.
+        """
+        super().__init__(**kwargs)
+        self._batch_count = {}
+        self._batch_count_saved = None
+        self._first_batch_in_epoch = True
+
+    def _reset(self) -> None:
+        """
+        Reset `_batch_count` and `_first_batch_in_epoch` to initial values.
```
petrbel: if only a oneliner, there's no need to make the docstring on a separate line

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+    .. code-block:: yaml
+        :caption: display progress of the current epoch
+
+        hooks:
+          - Progress
+    """
+
+    def __init__(self, **kwargs):
+        """
+        Create new Progress hook.
```
petrbel: if only a oneliner, there's no need to make the docstring on a separate line

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,119 @@
...
+Module with the hook which displays progress of the current epoch.
+"""
+
+import shutil
+
+from . import AbstractHook
+from ..datasets import AbstractDataset
+
+
+class Progress(AbstractHook):
```
blazekadam: Consider renaming the class to `ShowProgress`.

cxflow/hooks/progress.py
blazekadam: Consider using `click` package (http://click.pocoo.org/5/utils/#showing-progress-bars).

cxflow/hooks/progress.py
blazekadam: Not sure if one should call a method that can be overridden from constructor. Java does not like it for some reason...

cxflow/hooks/progress.py
gnovis: I tried to use it, but there were a lot of problems for using it in this hook, because there is not possible to use `click` progressbar as context manager.

cxflow/hooks/progress.py
```diff
@@ -0,0 +1,114 @@
...
+
+class ShowProgress(AbstractHook):
+    """
+    Show progress of a processed stream in the current epoch.
+
+    .. code-block:: yaml
+        :caption: show progress of the current epoch
+
+        hooks:
+          - Progress
```
blazekadam: Update the hook name in the example.

cxflow/hooks/progress.py
blazekadam: I see. I did not tried it, but cant you call the `__enter__` and `__exit__` methods manually? Anyways, your implementation looks nice. I am already using it.

gnovis: Yes it is possible, it solves one of the issue, but it seems like hack for me. Additionally there are another problems with removing progress bar and also it is not possible to solve situation when terminal window is smaller then progress bar line. So finally I decided to not to use it, because the solution was quite complicated.

PR #110, opened 2017-09-04
Downloadable dataset

petrbel:


cxflow/utils/download.py
```diff
@@ -0,0 +1,40 @@
...
+        filepath = os.path.join(data_root, filename)
+        if os.path.exists(filepath):
+            logging.info('\t`%s` already exists; skipping', filepath)
+            continue
+
+        os.makedirs(data_root, exist_ok=True)
+        logging.info('\tdownloading %s', filepath)
+        if not url_root.endswith('/'):
+            url_root += '/'
+        urllib.request.urlretrieve(url_root + filename, filepath)
```
blazekadam: Add `click` progress bar.

cxflow/utils/download.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if filename.endswith('.zip'):
```
blazekadam: Use `shutil`.

cxflow/datasets/downloadable_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+from abc import abstractmethod, ABCMeta
+from typing import Iterable
+
+from .base_dataset import BaseDataset
+from ..utils import maybe_download_and_extract
+
+
+class DownloadableDataset(BaseDataset, metaclass=ABCMeta):
+    """
+    Base class for datasets which download their data via :py:meth:`download`.
```
blazekadam: ``DownloadableDataset`` is dataset base class implementing routines for downloading and extracting data via ``cxflow dataset download`` command.

cxflow/datasets/downloadable_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+    @property
+    @abstractmethod
+    def download_filenames(self) -> Iterable[str]:
+        """A list of filenames to be downloaded from :py:meth:`url_root`."""
+        pass
+
+    def download(self) -> None:
+        """Maybe download and extra files required for training.
+
+        If not already donwloaded, download all files specified by :py:meth:`download_filenames` which are supposed to
```
blazekadam: do**wn**loaded

cxflow/datasets/downloadable_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+        be located in URL specified by :py:meth:`url_root`. Then, extract the downloaded files to :py:meth:`data_root`.
```
blazekadam: This sounds like the archives are kept somewhere else which is not true.

cxflow/datasets/downloadable_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+class DownloadableDataset(BaseDataset, metaclass=ABCMeta):
+    """
+    Base class for datasets which download their data via :py:meth:`download`.
+
+    In the inherited class, one should:
+        - implement :py:meth:`data_root` property
+        - implement :py:meth:`url_root` property
+        - implement :py:meth:`download_filenames` property
+    """
+
```
blazekadam: Consider implementing an `__init__` method which would require the three aforementioned parameters and save them in `self` for later use. 
The inherited class would pass `**kwargs` to `super().__init__` and stop caring about the download routines at all..

cxflow/utils/download.py
petrbel: done

PR #113, opened 2017-09-05
Progress ETA

gnovis:
Added estimated time to progress bar


blazekadam:
I think all this can be achieved more elegantly with `_after_epoch_profile`. Basically all the measurements needed for calculating ETA are already done by `main_loop` and provided as `epoch_profile` in the aforementioned event. You can sum the respective event times and obtain total stream processing time from it. You can even compensate for the first batch as `epoch_profile` contains timings of all batches including the first one.

I do not think that introducing this extra event is necessary, it would mean a lot of changes in docs etc..

gnovis:
Ok, I had the problem with determining the start of the stream in progress. Now it is solved without before_epoch event.

cxflow/hooks/progress.py
```diff
@@ -106,9 +127,10 @@ def after_batch(self, stream_name: str, batch_data: AbstractDataset.Batch):
...
         """
         After the first epoch the count of batches is saved, because of displaying the progress bar.
-        After every epoch the current batch count is reseted.
+        After every epoch the current batch count is reseted and start of the first stream in epoch is saved.
         """
         # after first epoch
         if not self._batch_count_saved:
             self._batch_count_saved = self._batch_count.copy()
         self._reset()
+        self._next_stream_start = time.time()
```
blazekadam: This is not a good place for start measuring the time as a number of things can be done between this event and the actual start of the next epoch, e.g.: other hooks may be called and do their stuff.

You can either
  - start the measurement in the `after_epoch_profile` event which is *closer* to the start of the next epoch
  - **(preferred)** start the measurement in the `after_batch` and compensate for the first batch according to the last `epoch_profile`. The epoch profile contains the timing of the 1st batch from the previous epoch which should be subtracted. You would need to capture the `epoch_profile` in `after_epoch_profile` event.

cxflow/hooks/progress.py
```diff
@@ -73,30 +82,42 @@ def after_batch(self, stream_name: str, batch_data: AbstractDataset.Batch):
...
         terminal_width = shutil.get_terminal_size().columns
 
         # not first epochs
         if self._batch_count_saved:
+
             total_batches = self._batch_count_saved[stream_name]
+
+            if stream_name in self._stream_duration:
+                # count the remaining time with respect to the previous epoch (for third and following epochs)
+                eta_sec = self._stream_duration[stream_name] - (time.time() - self._next_stream_start)
```
blazekadam: `self._next_stream_start` was not declared in `__init__`, also `_current_stream_start` would be a better name, would not it?

PR #114, opened 2017-09-06
Add AbstractHook, EpochData, Stream and Batch to cxflow API

blazekadam:


FloopCZ:
I am going to call you Mr. Exposer.

PR #132, opened 2017-09-20
Every n epoch base

gnovis:
Abstract `EveryNEpoch` hook, which call `after_n_epoch` method every `n_epochs` epoch.


cxflow/hooks/every_n_epoch.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, n_epochs: int=1, **kwargs):
+        """
+        Create EveryNEpoch hook.
+
+        :param n_epochs: how often ``after_n_epoch`` method is called
+        """
+        super().__init__(**kwargs)
+        self._n_epochs = n_epochs
+
+    def after_n_epoch(**kwargs):
```
petrbel: This should be renamed so that it is a private method.
This should also be an abstract method.

PR #152, opened 2017-10-05
Download also unextractable files.

petrbel:
When downloaded file cannot be extracted by shutil, just warn the user and continue instead of raising an exception.


cxflow/utils/download.py
```diff
@@ -41,4 +41,8 @@ def maybe_download_and_extract(data_root: str, url_root: str, filenames: Iterabl
                     f_out.flush()
 
         # extract
-        shutil.unpack_archive(filepath, data_root)
+        try:
+            shutil.unpack_archive(filepath, data_root)
+        except (shutil.ReadError, ValueError):
+            logging.info('File `%s` could not be extracted by `shutil.unpack_archive`. Please process it manually.',
```
blazekadam: This should be either `warning` or `error`. I would opt for `error`.

blazekadam: Apparently, this exception is thrown also for regular files; therefore, we should keep it on the `info` level...

PR #153, opened 2017-10-05
OnPlateau hook

gnovis:
Added new OnPlateau abstract hook from which StopOnPlateau is derived.


cxflow/hooks/on_plateau.py
```diff
@@ -0,0 +1,63 @@
...
+        self._loss_var = loss_var
+        self._saved_loss = []
+
+    @abstractmethod
+    def _on_plateau_action(self, **kwargs) -> None:
+        """
+        Abstract method which is called when the model stops improving.
+        This method must be overridden.
+
+        :raise TypeError: if this method is not overridden
```
blazekadam: Technically this method does not raise anything. I do not think we should include this annotation.

cxflow/hooks/on_plateau.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __init__(self, long_term: int=50, short_term: int=10, stream: str='test', loss_var: str='loss', **kwargs):
+        """
+        Create new OnPlateau hook.
+
+        :param long_term: count of last epochs representing long training period
+        :param short_term: count of last epochs representing short training period
+        :param stream: name of the processed stream
+        :param loss_var: name of the processed loss variable
+        :param kwargs: Ignored
```
blazekadam: Param descriptions should start with a small letter

cxflow/hooks/on_plateau.py
```diff
@@ -0,0 +1,63 @@
...
+    """
+
+    def __init__(self, long_term: int=50, short_term: int=10, stream: str='test', loss_var: str='loss', **kwargs):
+        """
+        Create new OnPlateau hook.
+
+        :param long_term: count of last epochs representing long training period
+        :param short_term: count of last epochs representing short training period
+        :param stream: name of the processed stream
+        :param loss_var: name of the processed loss variable
```
blazekadam: The variable being observed may be different to `loss`. We should  use more general name.

cxflow/hooks/on_plateau.py
```diff
@@ -0,0 +1,63 @@
+"""
+Module with OnPlateau abstract hook which call ``_on_plateau_action`` when the model stops improving.
+"""
+
+import numpy as np
+from abc import abstractmethod, ABCMeta
+
+from . import AbstractHook, ComputeStats
+
+_AGGREGATION = 'mean'
```
blazekadam: I would move this to `OnPlateau` class.

gnovis:
Thank you!

PR #157, opened 2017-10-07
Integrate CatchSigint to MainLoop, Remove CatchSigint hook

blazekadam:


cxflow/utils/misc.py
```diff
@@ -64,4 +66,61 @@ def __exit__(self, *args):
...
+    """
+
+    def __init__(self):
+        """Create new CatchSigint."""
+        self._num_sigints = 0
+        self._original_handler = None
+
+    def _sigint_handler(self, signum, _) -> None:
+        """
+        On the first signal, increase the ``self_num_sigints`` counter.
```
FloopCZ: `self._num_sigints`

cxflow/utils/misc.py
```diff
@@ -64,4 +66,61 @@ def __exit__(self, *args):
         logging.getLogger(self._name).disabled = self._orig
 
 
+class CatchSigint:
+    """
+    Catch SIGINT signals allowing to stop the training with grace in between ``__enter__`` and ``__exit__``.
+
+    On first sigint raise :py:class:`TrainingTerminated` in :py:meth:`raise_check_sigint`.
+    On second sigint quit immediately with exit code 1
```
FloopCZ: Missing dot at the end of the sentence (although Sphinx is not that sensitive to missing dots as Doxygen).

cxflow/utils/misc.py
```diff
@@ -64,4 +66,61 @@ def __exit__(self, *args):
...
+        if self._num_sigints > 0:  # not the first sigint
+            logging.error('Another SIGINT %s caught - terminating program immediately', signum)
+            quit(1)
+        else:  # first sigint
+            logging.warning('SIGINT %s caught', signum)
+            self._num_sigints += 1
+
+    def raise_check_sigint(self) -> None:
+        """
+        Stop the training if SIGINT signal was caught
```
FloopCZ: Missing dot at the end of the sentence.

cxflow/utils/misc.py
```diff
@@ -5,6 +5,8 @@
 import os
 import sys
 import logging
+import signal
+from ..hooks.abstract_hook import TrainingTerminated
```
FloopCZ: A few files above, you wrote just `from cxflow.hooks import TrainingTerminated`! :-)

cxflow/utils/misc.py
```diff
@@ -64,4 +66,61 @@ def __exit__(self, *args):
...
+            quit(1)
+        else:  # first sigint
+            logging.warning('SIGINT %s caught', signum)
+            self._num_sigints += 1
+
+    def raise_check_sigint(self) -> None:
+        """
+        Stop the training if SIGINT signal was caught
+
+        :raise TrainingTerminated: if an SIGINT signal was caught
```
FloopCZ: a SIGINT

cxflow/utils/misc.py
```diff
@@ -64,4 +66,61 @@ def __exit__(self, *args):
...
+        On the first signal, increase the ``self_num_sigints`` counter.
+        Quit on any subsequent signal.
+
+        :param signum: SIGINT signal number
+        """
+        if self._num_sigints > 0:  # not the first sigint
+            logging.error('Another SIGINT %s caught - terminating program immediately', signum)
+            quit(1)
+        else:  # first sigint
+            logging.warning('SIGINT %s caught', signum)
```
FloopCZ: We should probably provide some more informative message. E.g.,
```
logging.warning('SIGINT %s caught - training will be terminated after the current batch', signum)
```

cxflow/main_loop.py
```diff
@@ -116,6 +119,8 @@ def _run_epoch(self, stream: StreamWrapper, train: bool) -> None:
                 for hook in self._hooks:
                     hook.after_batch(stream_name=stream.name, batch_data={**batch_input, **batch_output})
 
+            self.raise_check_sigint()
```
FloopCZ: Calling this only here may be insufficient, because SIGINT from `before_training` and `after_epoch` would be ignored until the first batch of the next epoch and all the `after_batch_hooks` are finished processing. We should call it at the beginning and at the end of each epoch as well.

cxflow/utils/misc.py
blazekadam: We are in `CatchSigint`, we have no clue when the training will be terminated. Actually, we do not even know that we are training (although we talk about it in docs)...

blazekadam: We do not know when it will be terminated. It may not be terminated at all, it depends if `raise_check_sigint` is called, when it is called and how `TrainingTerminated` exception is handled.

cxflow/utils/misc.py
blazekadam: Here, we are *inside* the module while tests are *outside*, i.e., tests should use the regular public API.

cxflow/main_loop.py
blazekadam: Well, I am not sure about this. For example, what is happening between the last batch (and sigint check) and `after_epoch` after which you suggest to check sigint as well? In fact, only hooks are executed. So should we check for the sigint after all of them, or after every single one? Should we place the check between batch read and batch processing? We can check for it after every single statement but that is not the point right?

The point is to interrupt the training. Gracefully. I.e., to stop processing additional batches. I do not mind waiting for hooks to finish their work. You can send yet another sigint after all...

cxflow/utils/misc.py
blazekadam: `CatchSigint` does not control when it will be terminated. It depends on whether `raise_check_sigint` is called, when it is called and how `TrainingTerminated` is handled...

FloopCZ: I believe it will always be terminated. If we are inside the `CatchSigint` resource and catch a `SIGINT`, it is expected that `raise_check_sigint` will be called in the first convenient moment. It is a core part of `MainLoop` and even the exception is called `TrainingTerminated` so a developer somehow expects that it will terminate the training.

If you still prefer your message, maybe we could at least call it `Interrupt` instead of `SIGINT` when communicating with the users, since not everyone knows that `C^C` produces some kind of signals and that it is called `SIGINT`.

cxflow/main_loop.py
FloopCZ: > The point is to interrupt the training. Gracefully. I.e., to stop processing additional batches.

You are very right. That is why it is kind of unexpected if I press `C^C` during the `after_epoch` hooks that I have to wait for one more batch and the corresponding hooks. But yes, I surely expect to finish hook processing. In the very least, I would run the `raise_check` function __before__ processing the batch and not __after__.

Sending another `SIGINT` is not an option, since it may break my logs and all the data produced by the hooks. (Actually, I would prefer if there was a small (e.g., 1s) time limit between the registration of the first and the second `SIGINT` hook as it was in our older framework, because it really hurts to accidentally break many hours of computation.) Most of the UNIX programs are set up to never `quit()` on `C^C` and the users sometimes press the shortcut multiple times. I would even prefer if the message would inform that the next `C^C` will terminate immediately, similarly to `Blocks`.

blazekadam: I agree with moving the test before processing. Actually converged to the same idea while showering my hair :)

Regarding the delay... It never occurred to me and I am the infamous sudo force guy...

blazekadam: Now 
1st C^C would mean **do not process any more batches**
2nd C^C **kill it, no regards**

That sounds reasonable...

FloopCZ: Ok, I also like it this way (I am also talking about taking a shower in the morning :slightly_smiling_face: ).

PR #160, opened 2017-10-08
SaveLatest hook

gnovis:
Added `SaveLatest` hook with test. Fixed bug in `SaveBest` hook (`output_name` param was ignored). Removed unused imports.


cxflow/hooks/save.py
```diff
@@ -182,4 +182,37 @@ def after_epoch(self, epoch_data: EpochData, **_) -> None:
...
+        """
+
+        assert on_save_failure in SaveEvery.SAVE_FAILURE_ACTIONS
+
+        super().__init__(**kwargs)
+        self._model = model
+        self._output_name = output_name
+        self._on_save_failure = on_save_failure
+
+    def after_epoch(self, epoch_data: EpochData, **_) -> None:
```
blazekadam: Can we hide `epoch_data` to `**_`? It is not used, nor documented.

cxflow/hooks/save.py
```diff
@@ -182,4 +182,37 @@ def after_epoch(self, epoch_data: EpochData, **_) -> None:
...
+        """After every epoch overwrites the latest saved model with suffix `output_name`."""
```
blazekadam: This docstring is quite inaccurate. *latest saved model* may be understood as *globally latest*, e.g. saved by a different hook. Also it does not overwrite anything after the 1st epoch.

I would suggest:
```
Save/override the latest model after every epoch.
```

PR #161, opened 2017-10-08
Do not create logdir for `cxflow dataset` command

blazekadam:


blazekadam:
> I debug streams like this all the time
Right, we all **debug** streams like that. Do we really need to keep logs of these debug runs? If need be, you can dedicate a folder for some artifacts(images) like in this dataset method:
```python
    def visualize(self):
        """Dump first batches from both train and valid streams."""
        directory = 'visualize'
        if path.exists(directory):
            logging.info('Directory `%s` exists, cleaning up', directory)
            shutil.rmtree(directory)
        for stream_name in ['train', 'valid']:
            logging.info('Saving images from the first %s batch', stream_name)
            stream_dir = path.join(directory, stream_name)
            os.makedirs(stream_dir)
            for batch in getattr(self, '{}_stream'.format(stream_name))():
                for image, label, filepath in zip(batch['images'], batch['labels'], batch['filepaths']):
                    cv2.imwrite(path.join(stream_dir, path.basename(filepath)+'_'+str(label)+'.jpg'), image*255)
                break
```
Additionally, you do not have to
- find the folder in `log` dir
- clean up these folders

PR #162, opened 2017-10-08
Epoch Separator Fix

### LINKED ISSUE #159: fix logging epoch separator
In `LogVariables` hook `after_epoch` method are printed three new lines which should probably separates epochs logging, It seems like historical quick fix. I believe that epoch separator should be printed in a main loop, also "After Epoch ..." message should be removed from `LogVariables`.

gnovis:
Epoch separator moved from `LogVariables` to `MainLoop`
closes #159


blazekadam:
Can you explain your thinking behind this? In particular, I wonder why you removed the initial logging from `LogVariables`.

cxflow/main_loop.py
```diff
@@ -223,6 +223,7 @@ def training():
             # Zeroth epoch: after_epoch
             if not self._skip_zeroth_epoch:
                 self._run_zeroth_epoch([CXF_TRAIN_STREAM] + self._extra_streams)
+                logging.info('Zero epoch done\n\n')
```
blazekadam: These new lines were printed only to stderr. We do not want them to appear in the log file although the argument behind it is imo weak...

@FloopCZ  @petrbel whats your opinion?

cxflow/main_loop.py
```diff
@@ -237,14 +238,16 @@ def training():
...
+
                 with Timer('after_epoch_hooks', self._epoch_profile):
                     for hook in self._hooks:
                         hook.after_epoch(epoch_id=epoch_id, epoch_data=epoch_data)
 
                 for hook in self._hooks:
                     hook.after_epoch_profile(epoch_id=epoch_id, profile=self._epoch_profile,
                                              extra_streams=self._extra_streams)
-                logging.info('Epochs done: %s', epoch_id)
+                logging.info('Epochs done: %s\n\n', epoch_id)
```
blazekadam: ditto

cxflow/main_loop.py
FloopCZ: Yes, it is kind of strange to have two newlines at the end of a log message. If there _ever will be_ some kind of training management with a database where the logs are saved in separate records, than it would be messy. For now, we are stuck with `train.log` where it does not really matter.

petrbel: The original argument was readability of stderr. I think we should avoid blank lines in the saved log.

blazekadam: Because?

petrbel: because it contains no information, it's just visual

blazekadam: ooookay... I still do not see why we should avoid it. what is the negative effect? is there anything we ll have trouble with?

PR #163, opened 2017-10-10
Rework CatchSigint to CaughtInterrupts

blazekadam:
* catch both SIGINT and SIGTERM
* change 2nd signal handling from quit to sys.exit
* test 2nd signal handling
* remove redundant KeyboardInterrupt handling from MainLoop


cxflow/tests/utils/misc_test.py
```diff
@@ -53,35 +53,56 @@ def handler_set_event(signum, frame):
...
-        with CatchSigint():
-            kill(os.getpid(), signal.SIGINT)
-        self.assertFalse(self._sigint_unhandled, 'SigintHook does not handle SIGINT while training')
-        # double SIGINT cannot be easily tested (it quits(1))
+        """Test ``CatchInterrupts`` does handle interrupt signals inside with-resource environment."""
+        for sig in CatchInterrupts.INTERRUPT_SIGNALS:
+            with CatchInterrupts():
+                kill(os.getpid(), sig)
+            self.assertFalse(self._sigint_unhandled,
+                             'SigintHook does not handle signal `{}` inside with-resource environment'.format(sig))
```
FloopCZ: SigintHook -> CatchInterrupts

cxflow/utils/misc.py
```diff
@@ -66,62 +66,68 @@ def __exit__(self, *args):
...
+    INTERRUPT_SIGNALS = [signal.SIGINT, signal.SIGTERM]
+    """List of interrupt signals being handled by :py:class:`CatchInterrupts`."""
+
     def __init__(self):
         """Create new CatchSigint."""
-        self._num_sigints = 0
-        self._original_handler = None
+        self._num_signals = 0
+        self._origin_handlers = {}
+        self._original_sigint_handler = None
```
FloopCZ: remove, unused

cxflow/utils/misc.py
```diff
@@ -66,62 +66,68 @@ def __exit__(self, *args):
...
+        self._original_sigterm_handler = None
```
FloopCZ: ditto

cxflow/tests/utils/misc_test.py
```diff
@@ -53,35 +53,56 @@ def handler_set_event(signum, frame):
...
 
 
-class CatchSigintTest(CXTestCase):
-    """Test case for ``CatchSigint`` with-resource class."""
+class CatchInterruptsTest(CXTestCase):
+    """Test case for ``CatchInterrupts`` with-resource class."""
 
-    def _sigint_handler(self, *_):
+    def _interrupt_handler(self, *_):
         self._sigint_unhandled = True
```
FloopCZ: I suggest to rename this variable to `self._signal_unhandled`, since not only sigint is tested anymore

cxflow/utils/misc.py
```diff
@@ -66,62 +66,68 @@ def __exit__(self, *args):
...
-            catch.raise_check_sigint()  # raise TrainingTerminated if at least one sigint was caught
+            catch.raise_check_sigint()  # raise TrainingTerminated if at least one signal was caught
 
     """
 
+    INTERRUPT_SIGNALS = [signal.SIGINT, signal.SIGTERM]
+    """List of interrupt signals being handled by :py:class:`CatchInterrupts`."""
+
     def __init__(self):
         """Create new CatchSigint."""
```
FloopCZ: SigintHook -> CatchInterrupts

cxflow/utils/misc.py
```diff
@@ -66,62 +66,68 @@ def __exit__(self, *args):
...
-        :param signum: SIGINT signal number
+        On the first signal, increase the ``self._num_signals`` counter.
+        Call ``sys.exit`` on any subsequent signal.
         """
-        if self._num_sigints > 0:  # not the first sigint
+        if self._num_signals > 0:  # not the first signal
             logging.error('Another interrupt signal caught - terminating program immediately')
-            quit(1)
+            sys.exit(1)
         else:  # first sigint
```
FloopCZ: `# first signal`
(since you have fixed the comment three lines earlier, this one should be fixed too, otherwise it is confusing)

cxflow/utils/misc.py
```diff
@@ -66,62 +66,68 @@ def __exit__(self, *args):
...
     .. code-block:: python
         :caption: Usage
 
-        with CatchSigint() as catch:
-            # sigint signals are captured here
+        with CatchInterrupts() as catch:
+            # interrupt signals signals are captured here
             # do staff
-            catch.raise_check_sigint()  # raise TrainingTerminated if at least one sigint was caught
+            catch.raise_check_sigint()  # raise TrainingTerminated if at least one signal was caught
```
FloopCZ: Please rename to `raise_check_interrupt` or `raise_check_signal`, since not only SIGINT is caught anymore (you have actually renamed it in the comment, but not in the function name)

blazekadam: I do not agree with your proposal. The class is name `CaughtInterrupts` hence we check for interrupts. An interrupt may be either SIGINT or SIGTERM as both ultimately interrupt and terminate the training. I see the conflict with INT in SIGINT but I do not think that it is that much confusing. I would be confused by *signal* even more as it is quite general.

cxflow/utils/misc.py
```diff
@@ -66,62 +66,66 @@ def __exit__(self, *args):
...
-    On second sigint quit immediately with exit code 1.
+    On first signal raise :py:class:`TrainingTerminated` in :py:meth:`raise_check_interrupt`.
+    On second signal call ``sys.exit`` with exit status 1.
 
     .. code-block:: python
         :caption: Usage
 
-        with CatchSigint() as catch:
-            # sigint signals are captured here
+        with CaughthInterrupts() as catch:
```
FloopCZ: Caughth -> Caught

PR #165, opened 2017-10-11
Yaml/conf utils split

petrbel:


cxflow/utils/yaml.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def yaml_to_str(data: dict) -> str:
+    """
+    Return the given given config as YAML str.
+
+    :param data: configuration dict
+    :return: given configuration as yaml str
+    """
+    return yaml.dump(data, Dumper=ruamel.yaml.RoundTripDumper)
```
blazekadam: Missing newline.

cxflow/utils/yaml.py
```diff
@@ -0,0 +1,44 @@
...
+    Load YAML from file.
+
+    :param yaml_file: path to YAML file
+    :return: content of the YAML as dict/list
+    """
+    with open(yaml_file, 'r') as file:
+        return ruamel.yaml.load(file, ruamel.yaml.RoundTripLoader)
+
+
+def dump_to_yaml(data, output_dir: str, name: str) -> str:
```
blazekadam: Function name is inconsistent with `yaml_to_str`. Why not `yaml_to_file`? This is confusing, in this case `yaml` means a yaml file while in the second function `yaml` means a python object.

cxflow/utils/config.py
```diff
@@ -51,29 +48,4 @@ def load_config(config_file: str, additional_args: typing.Iterable[str]=()) -> d
     return config
 
 
-def config_to_file(config, output_dir: str, name: str=CXF_CONFIG_FILE) -> str:
```
blazekadam: What about forwarding these functions to `yaml...` counterparts instead of deleting them? It would make the diff much smaller...

cxflow/utils/yaml.py
petrbel: done

cxflow/utils/config.py
petrbel: consulted

PR #169, opened 2017-11-16
Remove type annotation from majority.py example

luxusko:
Provided type cx.AbstractDataset.Stream no longer exists in cxflow.


docs/tutorial.rst
```diff
@@ -94,12 +94,12 @@ To generate the *majority* data and provide the data streams we will implement a
             self._train_x, self._train_y = x[:int(.8 * n_examples)], y[:int(.8 * n_examples)]
             self._test_x, self._test_y = x[int(.8 * n_examples):], y[int(.8 * n_examples):]
 
-        def train_stream(self) -> cx.AbstractDataset.Stream:
+        def train_stream(self):
```
petrbel: Why can't we use `cx.Stream` instead?

docs/tutorial.rst
```diff
@@ -94,12 +94,12 @@ To generate the *majority* data and provide the data streams we will implement a
...
             for i in range(0, len(self._train_x), self.batch_size):
                 yield {'x': self._train_x[i: i + self.batch_size],
                        'y': self._train_y[i: i + self.batch_size]}
 
-        def test_stream(self) -> cx.AbstractDataset.Stream:
+        def test_stream(self):
```
petrbel: ditto

docs/tutorial.rst
blazekadam: We can.

blazekadam: And we should :)

PR #37, opened 2017-07-26
Add ls command to cxflow CLI

blazekadam:
## changes
- add ls command to cxflow CLI
- move `CXF_*` variables to a separate file
- include date and milliseconds in log entries
- log and config file names are now `CXF` constants
- add epochs done logging to cxflow `main_loop`

## `cxflow ls` docstring
```
    List training dirs contained in the given dir with options and outputs similar to the regular `ls` command.

    The function is accessible through cxflow CLI `cxflow ls`.

    :param dir_: dir to be listed
    :param recursive: walk recursively in sub-directories, stop at train dirs (--recursive option)
    :param all_: include train dirs with no epochs done (--all option)
    :param long: list more more details including net name, net and dataset class/modules,
                 age, duration and epochs done (--long option)
    :param verbose: print more verbose output with list of additional artifacts and training config,
                    applicable only when a single train dir si listed (--verbose option)
```

## `cxflow ls` examples
The following command examples are from my local isletnet project dir with
- log dir containing 100+ training dirs with the old log formats (i.e., the command can not derive epochs done, and timings; hence, they are included only when --all option is specified) and 1 training with the new log format
- one training dir with the new log format in the project root

Command:
```
cxflow ls
```
Output:
```
IsletNet_2017-07-26-00-34-11_2hous1d4
```

____

Command:
```
cxflow ls -la
```
Output:
```
total 134
IsletNetCNTK_2017-06-28-00-55-38__c6_r4dg  IsletNetCNTK  nets.IsletNetCNTK  datasets.IsletNetDataset  N/A        N/A        N/A
IsletNetCNTK_2017-06-28-00-56-55_3zbpx43s  IsletNetCNTK  nets.IsletNetCNTK  datasets.IsletNetDataset  N/A        N/A        N/A
IsletNetCNTK_2017-06-28-01-02-02_583kl5ug  IsletNetCNTK  nets.IsletNetCNTK  datasets.IsletNetDataset  N/A        N/A        N/A
IsletNetCNTK_2017-06-28-01-03-17_w44437u0  IsletNetCNTK  nets.IsletNetCNTK  datasets.IsletNetDataset  N/A        N/A        N/A
...
IsletNet_2017-07-25-22-59-21_1kqkqvu2      IsletNet      nets.IsletNet      datasets.IsletNetDataset  N/A        N/A        N/A
[...]

PR #172, opened 2017-11-17
Add cxflow ls command

blazekadam:
This is the original `cxflow ls` code modified to use `TrainingTrace` instead of log.

See #37 for examples.

### DIFF
--- a/cxflow/cli/ls.py
+++ b/cxflow/cli/ls.py
@@ -1,0 +1,233 @@
+"""
+Module with `cxflow ls` CLI function and associated util functions.
+"""
+
+import os
+import os.path as path
+from datetime import datetime
+from collections import defaultdict
+from typing import Iterable, Tuple, List
+
+from tabulate import tabulate
+from babel.dates import format_timedelta
+
+from ..utils import load_config, yaml_to_str
+from ..constants import CXF_DEFAULT_LOG_DIR, CXF_CONFIG_FILE, CXF_TRACE_FILE, CXF_NA_STR, CXF_LOG_FILE
+from ..utils import TrainingTrace, TrainingTraceKeys
+
+
+def print_boxed(str_: str) -> None:
+    """Print the given string in ASCII box."""
+    print(tabulate([[str_]], tablefmt='grid'))
+
+
+def path_total_size(path_: str) -> int:
+    """Compute total size of the given file/dir."""
+    if path.isfile(path_):
+        return path.getsize(path_)
+    total_size = 0
+    for root_dir, _, files in os.walk(path_):
+        for file_ in files:
+            total_size += path.getsize(path.join(root_dir, file_))
+    return total_size
+
+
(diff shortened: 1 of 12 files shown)


cxflow/cli/ls.py
```diff
@@ -0,0 +1,238 @@
...
+              'Consider specifying the directory to be listed as an argument.'.format(CXF_DEFAULT_LOG_DIR))
+        quit(1)
+
+    if not path.exists(dir_):
+        print('Specified dir `{}` does not exist'.format(dir_))
+        quit(1)
+
+    all_trainings = _ls_print_listing(dir_, recursive, all_, long)
+
+    if long and len(all_trainings) > 1:
```
FloopCZ: So if I specify long and there is only a single training, this quits quietly and outputs nothing? Maybe you wanted `or` instead of `and` ?

cxflow/cli/ls.py
```diff
@@ -0,0 +1,238 @@
...
+    artifacts = [(type_, name) + humanize_filesize(path_total_size(path.join(train_dir, name)))
+                 for type_, name in artifacts]
+    print(tabulate(artifacts, tablefmt='plain', floatfmt='3.1f'))
+    print()
+
+
+def list_train_dirs(dir_: str, recursive: bool, all_: bool, long: bool, verbose: bool) -> None:
+    """
+    List training dirs contained in the given dir with options and outputs similar to the regular `ls` command.
+    The function is accessible through cxflow_cls CLI `cxflow_cls ls`.
```
FloopCZ: I am probably just missing some knowledge, what is `cxflow_cls` please?

cxflow/cli/ls.py
```diff
@@ -0,0 +1,238 @@
...
+            age = format_timedelta(datetime.now() - start_datetime) + ' ago'
+            if end_datetime:
+                duration = format_timedelta(end_datetime - start_datetime)
+            else:
+                duration = CXF_NA_STR
+        else:
+            age = CXF_NA_STR
+            duration = CXF_NA_STR
+
+        epochs_done = trace[TrainingTraceKeys.EPOCHS_DONE] if trace[TrainingTraceKeys.EPOCHS_DONE] else 0
```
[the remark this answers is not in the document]
blazekadam: too late

cxflow/cli/ls.py
blazekadam: That is my very own **cxflow** CLI fork renamed so that it does not interfere with the regular **cxflow** :D

cxflow/cli/ls.py
blazekadam: With a single directory, there is nothing, to sum up really. One (long) line with all the information will be printed. The summary tables would be redundant.

FloopCZ: Good, thanks for clearing this out.

FloopCZ:
Great job today! Thanks for the changes, I am looking forward to update my **cxflow** installation.

PR #174, opened 2017-11-19
Create log dirs with correct permissions, Shorten log dir names

### LINKED ISSUE #170: Log dirs created by cxflow are inaccesible for group and others
By default, cxflow log dirs are created with
```
drwx------
```

Imo, it should be
```
drwxr-xr-x
```

Pls comment @FloopCZ, @petrbel ?

blazekadam:
Although small, this PR might be controversial.

Changes:
- create log dir with `os.makedirs` instead of `tempfile.mkdtemp` which sets dir permissions to `700`
- use new log dir name pattern `[DATE-TIME]_[MODEL-NAME]_[DOCKER_LIKE_RANDOM]`

(the original was `[MODEL-NAME]_[DATE-TIME]_[DOCKER_LIKE_RANDOM]_[RANDOM]`)

Thinking behind:
- moving datetime to the front is quite obvious, it allows intuitive chronological sorting
- docker like random rendered the tempfile random obsolete (I stopped using it entirely)
- new dir name is significantly shorter

FYI:
There are 14880 possible docker like random names. The chance of conflict (same model name, datetime and random) is therefore very low . Interestingly, with 140 names selected randomly there is roughly 50% chance of having a duplicity (birthday paradox). Perhaps we want to increase the number of names or introduce 3rd word (color?).

closes #170

### DIFF
--- a/cxflow/cli/common.py
+++ b/cxflow/cli/common.py
@@ -18,7 +18,7 @@ from ..utils.training_trace import TrainingTrace, TrainingTraceKeys
 from ..main_loop import MainLoop
 
 
-def create_output_dir(config: dict, output_root: str, default_model_name: str='NonameModel') -> str:
+def create_output_dir(config: dict, output_root: str, default_model_name: str='Unnamed') -> str:
     """
     Create output_dir under the given ``output_root`` and
         - dump the given config to YAML file under this dir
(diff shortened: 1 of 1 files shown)


cxflow/cli/common.py
```diff
@@ -42,9 +42,9 @@ def create_output_dir(config: dict, output_root: str, default_model_name: str='N
         logging.info('\tOutput root folder "%s" does not exist and will be created', output_root)
         os.makedirs(output_root)
 
-    output_dir = tempfile.mkdtemp(prefix='{}_{}_{}_'.format(model_name, datetime.now().strftime('%Y-%m-%d-%H-%M-%S'),
-                                                            get_random_name()),
-                                  dir=output_root)
+    output_dir = path.join(output_root, '{}_{}_{}'.format(datetime.now().strftime('%Y-%m-%d-%H-%M-%S'),
+                                                          model_name, get_random_name()))
+    os.makedirs(output_dir)
```
FloopCZ: When using `os.makedirs`, we should really make sure that a **new** directory is created and without race conditions. IMHO Relying on the probability is not good enough (especially with gridsearch and many simultaneous trainings, the probability of conflict is actually pretty high). We should not be sloppy.

blazekadam: What would you propose?

blazekadam: Yea, I had something similar in my mind. Pretty ugly though...

blazekadam: Although I cant think of any better way to do it. Graceful testing for existence would create race conditions... I'll implement it as you suggest.

blazekadam:
Updated so that the output dir is always created. Its hard to test this automatically though. I validated the code with some injections.

Regarding the dir naming, I see your point but I am still for it.

PR #175, opened 2017-11-19
Update EPOCHS_DONE in training trace after every epoch

### LINKED ISSUE #171: EpochsDone counter in TrainingTrace is updated only after the training ends
which makes `trace.yaml` file inconsistent with `train.log` during training. 

Furthermore, `ls` command incorrectly assumes no epoch was done for running training.

blazekadam:
closes #171


blazekadam:
I do not think so, it is not really an epoch. What should be the value anyways? 0, 1? Next, the model is not modified yet so I do not want to list this training unless `-a` is specified.

blazekadam:
Is it ready to be merged then?

PR #180, opened 2017-11-27
Allow mutable `batch_data`

petrbel:
Due to misimplementation, the `batch_data` passed among hooks is not mutable. In contrast, `epoch_data` are mutable just fine.

This PR unifies both behaviors.


blazekadam:
Very nice! I ll utilize it right away.

PR #150, opened 2017-10-02
Add NaN handling to ComputeStats hook

FloopCZ:
- Add nanmean, nanfraction and nancount aggregators.

PR #182, opened 2017-11-28
Allow logging nanmeans as well.

petrbel:
Since Cognexa/cxflow#150, it is impossible to write nanmeans to CSV training progress (only means are allowed).

This PR provides support for nanmeans as well which is crucial for many variables.


blazekadam:
+1 for the tests

petrbel:
I modified the streams subject to @FloopCZ's suggestion.

PR #185, opened 2017-11-30
Stop on NaN hooks.

### LINKED ISSUE #111: Implement additional hooks
Extend the set of hooks by e.g.

- [x] terminate when training is overfitted (this might be a base hook from which additional hooks will be derived (one e.g. terminates))
- [ ] terminate when some of the statistics is NaN

It would be nice to a have a `EveryNEpochs` hook, which would be a base hook. The purpose of this hook is to provide an API which allows its descendants to run some actions every e.g. `N` epochs. Some hooks already does this, to they should be reimplemented.

- [x] `EveryNEpochs`

petrbel:
- closes #111 

*(implemented while the id-ocr datasets are compiled, not wasting time :) )*


cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+            for variable in variables:
+                if variable not in stream_data:
+                    raise KeyError('Variable `{}` to be logged was not found in the batch data for stream `{}`. '
+                                   'Available variables are `{}`.'.format(variable, stream_name, stream_data.keys()))
+
+                value = stream_data[variable]
+                if StopOnNaN._is_nan(value):
+                    raise TrainingTerminated('Variable `{}` is NaN.'.format(variable))
+
+    def after_epoch(self, epoch_id: int, epoch_data: EpochData) -> None:
```
blazekadam: I would consider optional checking of `batch_data` in `after_batch` event. In some cases an epoch can take a long time to complete.

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        Stop the training once the epoch data contains a monitored variable equal to NaN.
+
+        :param epoch_id: number of processed epoch
```
blazekadam: unused variable may be hidden to `**kwargs` and does not have to be documented.

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+    .. code-block:: yaml
+        :caption: stop on NaN in loss variable
+
+        hooks:
+          - LogVariables:
+              variables: [loss]
+
+    """
+
+    def __init__(self, variables: Iterable[str]=None, **kwargs):
```
blazekadam: Can we support `inf` as well? E.g. an optional argument `stop_on_inf`? Although the name of the hook would be misleading.

cxflow/tests/hooks/stop_on_nan_test.py
```diff
@@ -0,0 +1,72 @@
+"""
+Module with stop on NaN hook test case (see :py:class:`cxflow.hooks.StopOnNan`).
+"""
+import numpy as np
+
+from cxflow.tests.test_core import CXTestCase
+from cxflow.hooks import StopOnNaN, TrainingTerminated
+
+
+class StopOnNaNTest(CXTestCase):
```
blazekadam: - `dict` recursion is not tested
- unsupported type is not tested
- only `np.nan` is tested, can you create some other NaNs?

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+        if isinstance(data, np.ndarray) or isinstance(data, list):
+            return any(np.isnan(data))
+        elif np.isscalar(data):
+            return np.isnan(data)
+        elif isinstance(data, dict):
+            return any([StopOnNaN._is_nan(value) for _, value in data.items()])
+        else:
+            raise ValueError('Unsupported is-NaN type `{}`.'.format(type(data)))
+
+    def _check_nan(self, epoch_data: EpochData):
```
blazekadam: missing return type annotation

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        Raise an exception when some of the monitored data is NaN.
+
+        :param epoch_data: epoch data checked
+        :raise KeyError: if the specified variable is not found in the stream
+        :raise TypeError: if the variable value is of unsupported type and ``self._on_unknown_type`` is set to ``error``
```
blazekadam: the description is deprecated (probably a copy-paste), at this moment the behavior is fixed to `error`

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        for stream_name in epoch_data.keys():
+            stream_data = epoch_data[stream_name]
+            variables = self._variables if self._variables is not None else stream_data.keys()
+            for variable in variables:
+                if variable not in stream_data:
+                    raise KeyError('Variable `{}` to be logged was not found in the batch data for stream `{}`. '
```
blazekadam: deprecated error message

cxflow/hooks/stop_on_nan.py
```diff
@@ -0,0 +1,86 @@
...
+        :caption: stop as soon as any variable contains NaN
+
+        hooks:
+          - StopOnNaN
+
+    .. code-block:: yaml
+        :caption: stop on NaN in loss variable
+
+        hooks:
+          - LogVariables:
```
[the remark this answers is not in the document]
luxusko: Why deprecated?

cxflow/hooks/stop_on_nan.py
luxusko: This is positional argument, is it possible to just remove it?

cxflow/tests/hooks/stop_on_nan_test.py
```diff
@@ -0,0 +1,150 @@
+"""
+Module with stop on NaN hook test case (see :py:class:`cxflow.hooks.StopOnNan`).
```
blazekadam: I can imagine writing this test with far less code. For example, consider the following function:
```python
def get_epoch_data(value):
    return {'simple_variable': value,
            'list_variable': [1, 2, value, 42],
            'dict_variable': {'key': value, 'key2': 84}}
```

now you would be able to cover most of the cases with
```python
for variable_name in ['simple_variable', 'list_variable', 'dict_variable']:
    StopOnNan(variables=[variable_name]).after_epoch(epoch_data=get_epoch_data(42))
    for bad_value in [np.nan, np.inf, float('nan'), float('inf'), float('-inf')]
        with self.assertRaises(TrainingTerminated):
             StopOnNan(variables=[variable_name], stop_on_inf=True).after_epoch(epoch_data=epoch_data(bad_value))
```

With this, you would be able to delete like 60 lines...

**warning: the code is not tested**

cxflow/tests/hooks/stop_on_nan_test.py
```diff
@@ -0,0 +1,150 @@
...
+                'valid': {'variable1': 2, 'variable2': [1, 2, 3]},
+            })
+
+        StopOnNaN().after_epoch(1, {
+            'train': {
+                'variable1': {'nested1': {'nested2': {'nested3': 0}, 'var2': 0}, 'var1': 0},
+                'variable2': [1, 2, 3]
+            },
+            'valid': {'variable1': 2, 'variable2': [1, 2, 3]},
+        })
```
blazekadam: Not tested:
- check nans in `after_batch` event if and only if `after_batch=True`.
- `KeyError` raised when a variable is specified but missing

blazekadam: See https://coveralls.io/builds/14572007/source?filename=cxflow%2Fhooks%2Fstop_on_nan.py

Issue #166, opened 2017-10-15
StreamWrapper re-implementation


blazekadam:
Current implementation of `StreamWrapper` is problematic with `buffer`>0. Up to date the following issues were encountered:

- T train increases as T read data decreases (which questions if there is any gain from the buffering)
- stopping and resuming an epoch with cxstream dataset sometimes causes segaults (minimal working example is not available, the problem disappeared with `buffer=0`)

It is suggested to re-implement `StreamWrapper` with either `asyncio` or python `concurrent` futures.

petrbel:
@Teyras afaik mentioned gevent?

Teyras:
I have no idea what are you guys talking about, but asyncio and gevent are both suited for IO-bound tasks. For CPU-bound tasks, you're better off using concurrent.

blazekadam:
I finally had some time to investigate this issue. Apparently, `StreamWrapper` works as expected; nonetheless, the time measurements are confusing. 

In short, I was able to save roughly 70% of epoch wall-clock time which could be potentially saved.

In particular, I carried out an experiment with one of our projects which utilizes **cxtream**. The following table captures [T Read] / [T Train] ([Epoch wall-clock]) values for different buffer set ups.

|   | cxflow no buffer | cxflow buffer=4  |
|---------------------------------|------------------|------------------|
| **cxtream no buffer**               | 46/23 (69)       | 1.6/52 (53)      |
| **cxtream buffer = 4**             | 0.6/23 (24)      | 0.03 / 24.5 (25) |

We can see that without **cxtream** buffer, TRead is greater than TTrain, hence we should be able to save some time with GPU-CPU parallelism, 23 seconds in particular. Note that this saving is only theoretical as TTrain includes the whole `mode.run` method call rather than pure GPU time.

Apparently, **cxflow** buffer indeed saves fair amount of time. More precisely, 16 seconds per epoch which is roughly 70% of the theoretical saving. Not bad huh?

Now the counter-intuitive thing here is that TTrain was larger despite wall clock time being smaller. With some logging in both `StreamWrapper` and `Model.run` method, I discovered that with **cxflow** buffer,  `model.run` method is being interrupted at different places for computing new batches. Those interruptions are of course included in TTrain.

Given all that, it is questionable whether we want to change the implementation and if so, how? @FloopCZ  @petrbel any comments?

FloopCZ:
[...]

PR #199, opened 2017-12-18
Add precise buffer management

FloopCZ:
Relates to #166.


cxflow/types.py
```diff
@@ -12,3 +12,7 @@
 
 TimeProfile = Mapping[str, List[float]]
 """Time profile type."""
+
+class TrainingTerminated(Exception):
```
blazekadam: I like moving the exception definition here.

cxflow/datasets/stream_wrapper.py
```diff
@@ -207,32 +221,73 @@ def __enter__(self) -> Iterator[Batch]:
...
+            :caption: Usage
+
+            # the training method of a model
+            def run(self, batch, train, stream):
+                preprocess_batch_in_python(batch)  # this function holds the GIL and fully utilizes the CPU
+                with stream.allow_buffering:
+                    call_native_backend(batch)  # this function is blocking, but releases the GIL
+                                                # we can use the GIL and the spare CPU to prepare the next batch
+
+        :return: A :py:class:`ReleasedLock` resource.
```
blazekadam: You do not have to craft links like this for return value docs. Sphinx automatically places the type link next to your text if you annotate the return type with python `typing`.

cxflow/datasets/stream_wrapper.py
```diff
@@ -102,7 +108,15 @@ def _enqueue_batches(self, stop_event: Event) -> None:
         :param stop_event: event signaling stop instruction
         """
         while True:
-            for batch in self._get_stream():
+            self._stream = self._get_stream()
+            while True:
+                self._lock.acquire()
+                self._lock.release()
+                time.sleep(CXF_BUFFER_SLEEP)
```
blazekadam: A comment with some reasoning behind this sleep can be helpful later.

cxflow/datasets/stream_wrapper.py
```diff
@@ -207,32 +221,73 @@ def __enter__(self) -> Iterator[Batch]:
...
+        # get the next batch and if the buffer is empty, allow buffering
+        def get_batch_maybe_buffer():
+            # buffering is fully disabled; just compute the next batch
+            if self._buffer_size <= 0:
+                return self._next_batch()
+            # something is in the buffer; get it
+            if not self.empty():
+                return self._dequeue_batch()
+            # the buffer is empty; allow buffering and wait
+            with self.allow_buffering:
```
blazekadam: Is this actually safe? I mean can you nest `with self.allow_buffering` because it can be called already. Perhaps we want to test it.

cxflow/constants.py
```diff
@@ -33,6 +33,9 @@
 CXF_NA_STR = 'N/A'
 """N/A string for pretty printing."""
 
+CXF_BUFFER_SLEEP = 0.02
```
petrbel: The constant is not configurable by the user. We should introduce some configuration file, e.g. `~/.config/cxflow`, that would overwrite these constants

cxflow/datasets/__init__.py
```diff
@@ -3,9 +3,11 @@
 from .abstract_dataset import AbstractDataset
 from .base_dataset import BaseDataset
 from .downloadable_dataset import DownloadableDataset
+from .stream_wrapper import StreamWrapper
 
 AbstractDataset.__module__ = '.datasets'
 BaseDataset.__module__ = '.datasets'
 DownloadableDataset.__module__ = '.datasets'
+StreamWrapper.__module__ = '.datasets'
```
petrbel: this is getting ugly :-/ should replace it with a loop?

cxflow/models/abstract_model.py
```diff
@@ -43,14 +43,15 @@ def output_names(self) -> Iterable[str]:
         pass
 
     @abstractmethod
-    def run(self, batch: Batch, train: bool) -> Batch:
+    def run(self, batch: Batch, train: bool, stream: StreamWrapper) -> Batch:
```
petrbel: I think it's a bit weird to pass both batch and stream (since batch can be obtained from stream, doesn't it?).

blazekadam: I am ok with it. It actually opens new possibilities.

cxflow/datasets/__init__.py
blazekadam: I would leave it like this. Having a loop in an init file seems wrong to me for some reason...

cxflow/constants.py
blazekadam: I had the same thoughts. On the other hand, I am not sure if the user is actually interested in configuring any of our constants at this moment. We ll introduce the configuration file once we integrate the experiment management to the core so I would leave it like this for now.

This PR is more about patching the `StreamWrapper` than some conceptual changes.

blazekadam:
With additional testing, it appears to be working well.

cxflow/datasets/stream_wrapper.py
FloopCZ: Indeed.

cxflow/datasets/stream_wrapper.py
FloopCZ: Good point, nesting will not work, it should be handled somehow.

cxflow/constants.py
FloopCZ: The global config is a good idea, but I agree with @blazekadam, that this is not necessary at the moment. Plus any reasonable time delay will improve upon the current state.

cxflow/models/abstract_model.py
FloopCZ: This is an interesting idea to only pass the stream. Unfortunately, we need the batch in the main loop as well, which would not be easy if the user would extract the batch out of the stream by himself in the `run` function.

FloopCZ:
Thank you for the thorough review, the requests have been addressed.

blazekadam:
Merging as the aforementioned issued is most likely related to some OpenCV stuff in only one of our streams.

PR #204, opened 2018-02-11
Add EnsembleModel including tests and docs

### LINKED ISSUE #15: Ensembles
Research the possibility of having `cxflow ensemble` subcommand that would merge multiple computationals graph into one and dumping a regular checkpoint.

@petrbel :
Just to sum up the discussion we had last week
- we need to benchmark few nets (various number of params, rnn/cnn, ..) in order to determine how fast/slow is the
   - decoding the network from file
   - loading the network to the GPU
- then we should determine whether manualy force the computational graphs to merge somehow or simple evaluate multiple networks serially

I prefer the later option, hence I suggest creating both cxtream and fuel NetApplier of EnsembleApplier.

blazekadam:
Easy usage from python
```python
model = cx.models.EnsembleModel(inputs=['images'], outputs=['predictions'], 
                                models_root='/my/directory/with/models')
```

For details, see the docs.

closes #15

### DIFF
--- a/cxflow/models/ensemble_model.py
+++ b/cxflow/models/ensemble_model.py
@@ -1,0 +1,186 @@
+import logging
+import os
+import os.path as path
+from typing import Iterable, Optional, Sequence
+from collections import Counter
+
+import numpy as np
+
+from .abstract_model import AbstractModel
+from ..utils import load_config
+from ..cli.common import create_model
+from ..datasets import AbstractDataset, StreamWrapper
+from ..types import Batch
+from ..constants import CXF_CONFIG_FILE
+
+
+def major_vote(all_votes: Iterable[Iterable[object]]) -> Iterable[object]:
+    """
+    For the given iterable of object iterations, return an iterable of the most common object at each position of the
+    inner iterations.
+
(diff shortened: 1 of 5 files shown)


cxflow/models/ensemble_model.py
```diff
@@ -0,0 +1,184 @@
...
+        self._models = None
+        if eager_loading:
+            self._load_models()
+
+    def _load_models(self) -> None:
+        """Maybe load all the models to be assembled together and save them to the ``self._models`` attribute."""
+        if self._models is None:
+            logging.info('Loading %s models', len(self._model_paths))
+
+            def load_model(model_path: str):
```
petrbel: Consider moving the nested function directly into the class

cxflow/models/ensemble_model.py
```diff
@@ -0,0 +1,184 @@
...
+    def save(self, *args, **kwargs) -> None:
+        """
+        Ensemble model can not be saved.
+
+        :raise NotImplementedError: when called
+        """
+        raise NotImplementedError('Ensemble model can not be saved.')
+
+    @property
+    def restore_fallback(self) -> None:
```
petrbel: this should raise an exception rather than return `None` instead of expected `str`

PR #207, opened 2018-02-21
Add confusion matrix hook

bvavrik:


cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,134 @@
+import numpy as np
+import os.path as path
+
+import itertools
+from sklearn.metrics import confusion_matrix
```
petrbel: `scikit` is a huge dependency. I'm really not sure we want/should use it in pure cxflow. Implementing a simple confussion matrix is basically just double defaultdict and a for-loop, isn't it?

Btw this dependency is missing from requirements file...

@blazekadam @FloopCZ ?

blazekadam: Yea, I am not sure either especially if it's just for the `confusion_matrix`. That should be like 5 lines.

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+    FIGURE_ACTIONS = ['save', 'store']
+    """
+    Possible actions to be taken with the plotted figure. 
+    It can be either saved to a file or stored in the epoch data.
+    """
+
+    def __init__(self,
+                 output_dir: str,
+                 dataset: BaseDataset,
+                 annotations_name: str='annotations',
```
petrbel: `annotations` is a quite surprising default for me since we've been (to my knowledge) using rather `labels` ever since.

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+                 probabilities_name: str= 'probabilities',
+                 classes_name: str=None,
```
petrbel: missing docstring

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+        :param cmap: type of colorbar  # http://matplotlib.org/examples/color/colormaps_reference.html
+        """
+        if figure_action not in SaveConfusionMatrix.FIGURE_ACTIONS:
+            raise ValueError('Unrecognized figure action `{}`. It must be one of `{}`'.
+                             format(figure_action, SaveConfusionMatrix.FIGURE_ACTIONS))
+
+        self._dataset = dataset
+        self._output_dir = output_dir
+        self._annotations_name = annotations_name
+        self._probabilities_name = probabilities_name
```
petrbel: Btw why probabilities? It would be easier to fetch predictions directly from the model, wouldn't it? AFAIK all models implement it...

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+        self._classes_name = classes_name
+        self._figure_action = figure_action
+        self._normalize = normalize
+        self._cmap = cmap
+
+        super().__init__(variables=list(filter(None, [annotations_name, probabilities_name, classes_name])), **kwargs)
```
petrbel: cool usage of `filter`, however, both `annotations_name` and `probabilities_name` can never be None or an empty string (otherwise the behavior of the hook would be destroyed).

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def after_epoch(self, epoch_id: int, epoch_data: EpochData) -> None:
+        for stream_name, variables in self._accumulator.items():
+
+            probabilities = np.array(variables[self._probabilities_name])
+            predictions = np.argmax(probabilities, axis=-1)
```
petrbel: see comment above - predictions are usually provided directly by the model (computed on GPU)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+            annotations = np.array(variables[self._annotations_name])
+            # If annotations or have shape `num_data*num_classes`, take only the index of annotated class
+            if len(annotations.shape) == 2:
```
petrbel: you can use `ndim` instead of `len(...shape)`

petrbel: I'd check for higher ndims and raise an exception on unexpected formats of annotations

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+                annotations = np.argmax(annotations, axis=-1)
+
+            # Get number of classes from `probabilities` shape
+            tick_marks = np.arange(probabilities.shape[-1])
```
petrbel: the code would be more readable if we (or preferably the dataset) had a variable/property `num_classes` that could be used here and in the `cm` creation

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+            # Save the heatmap of confusion matrix
+            plt.imshow(cm, interpolation='nearest', cmap=self._cmap)
+            plt.title('Confusion matrix')
+            plt.ylabel('Expected')
+            plt.xlabel('Predicted')
+            plt.colorbar()
+
+            if self._classes_name is not None:
+                # Set default name of classes to `NA`
+                classes = np.array(["NA"]*len(tick_marks))
```
petrbel: `"` -> `'`

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+                # When a class is present in data, add its name
+                for annotation in set(annotations):
+                    classes[annotation] = np.array(variables[self._classes_name])[list(annotations).index(annotation)]
```
petrbel: this should be obtained from the dataset

petrbel: btw why do we cast np array `annotations` to `list` over and over again in the for-loop?
too slow

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+            # Save / store the figure
+            if self._figure_action == 'store':
+                fig = plt.gcf()
+                # Draw the figure first
+                fig.canvas.draw()
+                # Now we can save it to a numpy array
+                data = np.fromstring(fig.canvas.tostring_rgb(), dtype=np.uint8, sep='')
+                data = data.reshape(fig.canvas.get_width_height()[::-1] + (3,))
+                epoch_data[stream_name]['confusion_heatmap'] = data
+            else:
```
petrbel: This should be `elif`... then there should be added `else` branch raising an exception on invalid action. This should be btw checked in the constructor.

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,136 @@
...
+import matplotlib.pyplot as plt
+
+from . import AccumulateVariables
+from ..types import EpochData
+from ..datasets import BaseDataset
+
+
+class SaveConfusionMatrix(AccumulateVariables):
+    """
+    Compute and save conufusion matrix between the predicted and expected labels.
```
petrbel: and -> or
between -> based on

cxflow/hooks/confusion_matrix.py
petrbel: this should be obtained from dataset, not from the stream, so I suggest removing this variable entirely

cxflow/hooks/confusion_matrix.py
bvavrik: Yeah, *we* are . But `cxflow` should by an open library, so we should use here some common defaults, no clear idea, which is better.

cxflow/hooks/confusion_matrix.py
bvavrik: `predictions` are tricky, because they can be `zeros` for some inputs. What would then confusion matrix do?
We can add a special `blank_class`, so if the `prediction` is zero for every class, we'll add points to `blank_class`

cxflow/hooks/confusion_matrix.py
bvavrik: That `filter` is used for `classes_name` only, will be deleted after `classes_name` removal

cxflow/hooks/confusion_matrix.py
bvavrik: see comment above :slightly_smiling_face:

cxflow/hooks/confusion_matrix.py
bvavrik: `index()` method is not provided to `np.ndarray`. I can use `np.where`, but this seems more readable for me

cxflow/hooks/confusion_matrix.py
bvavrik: I have there an error raise up in `__init__`:
```
        if figure_action not in SaveConfusionMatrix.FIGURE_ACTIONS:
            raise ValueError('Unrecognized figure action `{}`. It must be one of `{}`'.
                             format(figure_action, SaveConfusionMatrix.FIGURE_ACTIONS))
```

cxflow/hooks/confusion_matrix.py
petrbel: it's ok, I don't insist :) maybe we shouldn't use any default? weird...

cxflow/hooks/confusion_matrix.py
petrbel: Hmmm, I don't understand - the model predicts something (btw usualy via argmax as you do). One of the predictable classes can be "blank" it the task/dataset requires that. Nevertheless, this hook should work with the model prediction only (regardless of its meaning), shouldn't it?

cxflow/hooks/confusion_matrix.py
petrbel: The thing is we don't need it, do we? This can be easily provided by the dataset.

cxflow/hooks/confusion_matrix.py
petrbel: oh, my bad, sorry :)

cxflow/hooks/confusion_matrix.py
bvavrik: I would let you guys say what's the best option, you've been through more libraries than I :smile:

cxflow/hooks/confusion_matrix.py
bvavrik: Usually we just `round(probabilities)` (at least in CATI), what can cause the problem. I barely remember some discussion about choosing between `argmax` and `round` and the reason for choosing `round` is that we don't want to assume in accuracy some random good predictions

cxflow/hooks/confusion_matrix.py
bvavrik: Oh yeah, of course, if we'll fix the problem with classes names, I'm totally in to delete this part of code personally :smile:

cxflow/hooks/confusion_matrix.py
bvavrik: What is used more by ML people, `round` or `argmax`?
Defintely our hook should be robust in each case, so using `probabilities` is a save bet

petrbel: Maybe @blazekadam and @FloopCZ could give us more insight. 

AFAIK, in multi-categorical classification one usually uses argmax. Each potential class has its own (softmaxed) number and we're looking for the greatest.

In binary classification, one can also employ only one output number/channel (e.g. sigmoided) representing the final probability of class 1. Then, rounding makes sense.

Anyway, it's still the model's business. The hook doesn't even have to know the implementation of its prediction, does it?

bvavrik: " The hook doesn't even have to know the implementation of its prediction"
That's exactly why I've chosen `probabilities`, they are working for each case.

petrbel: But so do predictions, don't they? I really don't see your point, I'm sorry if I'm retarted :D

bvavrik: if anyone is having `predictions=round(probabilities)` in model, loading `predictions` can cause having zero vector as a row. So confusion matrix cannot find the predicted class (cause they are just zeros all of them)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        self._normalize = normalize
+        self._cmap = cmap
+
+        super().__init__(variables=list([labels_name, predictions_name]), **kwargs)
+
+    def after_epoch(self, epoch_id: int, epoch_data: EpochData) -> None:
+        for stream_name, variables in self._accumulator.items():
+            predicted = np.array(variables[self._predictions_name])
+            expected = np.array(variables[self._labels_name])
+            # If annotations or have shape `num_data*num_classes`, take only the index of annotated class
```
blazekadam: I do not think we need to deal with probs like this. Btw, we will rarely see one-hot encoded `expected` variables. From my pov, we should support either:
- fully general object (numbers, strings, tuples, whatever)
- numbers (class indices) only

In both options, however, we should treat both `expected` and `predicted` variables in the same way. After some thoughts, I would opt for the second option for its simplicity.

I.e., just assert `ndim==1` and int dtype and state the limited support in the docstring.

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                try:
+                    classes_names = getattr(self._dataset, self._classes_names_method_name)()
+                except AttributeError:
+                    pass
+            # Change ticks if `classes_names` found
+            if classes_names:
+                plt.xticks(np.arange(num_classes), classes_names, rotation=90)
+                plt.yticks(np.arange(num_classes), classes_names)
+
+            # Add both normalized and absolute values to graph
```
blazekadam: cool :)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            thresh = np.nanmax(cm) / 2.
```
blazekadam: even cooler :)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            for i, j in itertools.product(range(cm.shape[0]), range(cm.shape[1])):
+                plt.text(j, i, '{} / {}'.format(format(cm_norm[i, j], '.2f'), cm_abs[i, j]), fontdict={'size': 8},
```
blazekadam: I think you want to format with `{:.2f}` instead of this nesting.

cxflow/hooks/confusion_matrix.py
blazekadam: Just explain the threshold with docstring pls :)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    .. code-block:: yaml
+        :caption: Store confusion matrix heatmap
+
+        hooks:
+          - SaveConfusionMatrix:
+              figure_action: store
+
+    .. code-block:: yaml
+        :caption: Save confusion matrix heatmap with absolute values
```
blazekadam: *Save confusion matrix to training logdir with ...*

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+        hooks:
+          - SaveConfusionMatrix:
+              normalize: False
```
blazekadam: Showing a few more args in the examples (with docstings?) would be nice...

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self,
+                 output_dir: str,
+                 dataset: BaseDataset,
+                 labels_name: str= 'labels',
+                 predictions_name: str= 'predictions',
+                 classes_names: List[str]=None,
+                 figure_action: str='save',
+                 num_classes_method_name: str='num_classes',
+                 classes_names_method_name: str='classes_name',
+                 save_cm=False,
```
blazekadam: missing type annotation

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                 normalize=True,
```
blazekadam: ditto

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        :param predictions_name: prediction variable name
+        :param classes_names: List of classes' names
+        :param figure_action: action to be taken with the plotted figure, one of :py:attr:`FIGURE_ACTIONS`
+        :param normalize: False for plotting absolute values in confusion matrix, True for relative
+        :param num_classes_method_name: `self._dataset` method name to get number of classes
+        :param classes_names_method_name: `self._dataset` method name to get classes' names
+                                            Parameter is ignored when `classes_names` is provided
+        :param save_cm: True if confusion matrix values has to be stored to `epoch_data`, False otherwise
+        :param cmap: type of colorbar  # http://matplotlib.org/examples/color/colormaps_reference.html
+        """
```
blazekadam: Missing `:raise ValueError: if ...` docstring

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+    Possible actions to be taken with the plotted figure. 
+    It can be either saved to a file or stored in the epoch data.
+    """
+
+    def __init__(self,
+                 output_dir: str,
+                 dataset: BaseDataset,
+                 labels_name: str= 'labels',
+                 predictions_name: str= 'predictions',
+                 classes_names: List[str]=None,
```
blazekadam: This is actually `Optional[Sequence[str]]`

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            # Try to get names of classes from possible sources
+            if self._classes_names is not None:
+                classes_names = self._classes_names
+            else:
+                try:
+                    classes_names = getattr(self._dataset, self._classes_names_method_name)()
+                except AttributeError:
+                    pass
+            # Change ticks if `classes_names` found
+            if classes_names:
```
blazekadam: This will cause an error if `self._classes_names` is `None` and the dataset does not provide the class names (variable `classes_names` will not be declared!)

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                plt.xticks(np.arange(num_classes), classes_names, rotation=90)
```
blazekadam: It would be lovely to have the x-ticks at the top, that is much easier to read...

cxflow/hooks/confusion_matrix.py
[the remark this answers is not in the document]
bvavrik: Well, this is not an action with figure, but with `cm` itself.
The question is, if we want to save the figure as an image to `epoch_data` in any case (as it is in https://github.com/Cognexa/id-ocr/blob/dev/ocr/hooks.py). Wouldn't it be sufficient to have just two options: saving figure as `png` and saving data from `cm` to `epoch_data`?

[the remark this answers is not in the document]
blazekadam: I do realize the difference. Nonetheless, it is confusing that putting something to `epoch_data` is *saving* in one case and *storing* in another.

We want to support saving the image to `epoch_data` so that it can be displayed in the tensorboard.

If anything, we do not need to support `store_cm` option. It was my idea how to make the whole hook testable. You can make (possibly static) `compute_cm` method which would return both absolute and relative cms and test this function.

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        Create new :py:class:`SaveConfusionMatrix` hook.
+        :param output_dir: output directory
+        :param dataset: dataset (needed to translate predictions to strings)
+        :param labels_name: annotation variable name
+        :param predictions_name: prediction variable name
+        :param classes_names: List of classes' names
+        :param figure_action: action to be taken with the plotted figure, one of :py:attr:`FIGURE_ACTIONS`
+        :param normalize: False for plotting absolute values in confusion matrix, True for relative
+        :param num_classes_method_name: `self._dataset` method name to get number of classes
```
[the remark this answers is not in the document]
bvavrik: On the other hand, hardcoding those args is reasonable?
We give as an option to choose name of `predictions` and `labels`, I would stay consistent about giving the option here as well

cxflow/hooks/confusion_matrix.py
```diff
@@ -0,0 +1,155 @@
...
+    FIGURE_ACTIONS = ['save', 'store']
+    """
+    Possible actions to be taken with the plotted figure. 
+    It can be either saved to a file or stored in the epoch data.
+    """
+
+    def __init__(self,
+                 output_dir: str,
+                 dataset: BaseDataset,
+                 labels_name: str= 'labels',
```
petrbel: spaces only on the right side of = is pretty weird

blazekadam: We usually opt for no spaces around `=` in this case.

blazekadam:
@petrbel 
image

cxflow/tests/hooks/save_cm_test.py
```diff
@@ -0,0 +1,86 @@
+import os
+import collections
+
+from cxflow.tests.test_core import CXTestCaseWithDir
+from cxflow.hooks.save_cm import SaveConfusionMatrix
+from ..main_loop_test import SimpleDataset
+
+
+def _get_epoch_data():
+    epoch_data = collections.OrderedDict([
```
blazekadam: Not sure if you need `OrderedDict`s here but it will not hurt us

cxflow/tests/hooks/save_cm_test.py
```diff
@@ -0,0 +1,86 @@
...
+            SaveConfusionMatrixTest.run_hook(hook)
+        with self.assertRaises(ValueError):
+            hook = SaveConfusionMatrix(dataset=TestDataset(), output_dir=self.tmpdir,
+                                       cmap='fake')
+            SaveConfusionMatrixTest.run_hook(hook)
+        with self.assertRaises(ValueError):
+            hook = SaveConfusionMatrix(dataset=TestDataset(), output_dir=self.tmpdir,
+                                       figure_action='non_existing')
+        # test wrong number of classes' names
+            SaveConfusionMatrixTest.run_hook(hook)
```
blazekadam: This line is apparently not executed. The reason is that the preceding statement raises an exception an the `with ...` environment is therefore exited.

cxflow/tests/hooks/save_cm_test.py
```diff
@@ -0,0 +1,86 @@
...
+    epoch_data = collections.OrderedDict([
+        ('train', collections.OrderedDict([
+            ('accuracy', 1)]))])
+    return epoch_data
+
+
+class TestDataset(SimpleDataset):
+
+    @staticmethod
+    def num_classes():
```
blazekadam: This method is never utilized. Either use it in the tests or remove it.

cxflow/tests/utils/confusion_matrix_test.py
```diff
@@ -0,0 +1,57 @@
...
+from cxflow.utils.confusion_matrix import confusion_matrix
+from cxflow.tests.test_core import CXTestCaseWithDir
+
+
+class ConfusionMatrixTest(CXTestCaseWithDir):
+    """Test case for the confusion_matrix."""
+
+    def test_confusion_matrix(self):
+        # test insufficient arguments
+        self.assertRaises(TypeError, confusion_matrix)
```
blazekadam: This is very thorough! Usually, we do not test *insufficient arguments* as it is a feature of the language, not ours :)

cxflow/tests/utils/confusion_matrix_test.py
```diff
@@ -0,0 +1,57 @@
...
+        num_classes = 3
+        calculated_cm = confusion_matrix(expected=expected, predicted=predicted, num_classes=num_classes)
+        groundtruth_cm = np.array([[1, 0, 0],
+                                   [0, 1, 0],
+                                   [0, 1, 0]])
+        np.testing.assert_equal(calculated_cm, groundtruth_cm)
+
+        num_classes = 4
+        calculated_cm = confusion_matrix(expected=expected, predicted=predicted, num_classes=num_classes)
+        groundtruth_cm = np.array([[1, 0, 0, 0],
```
blazekadam: I would like to see more complex cms. E.g. with at least one value greater than one. You usually want to cover all the imaginable inputs and outputs. All your test cases are in fact binary matrices which looks like one special case.

cxflow/utils/confusion_matrix.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+
+def confusion_matrix(expected: np.ndarray, predicted: np.ndarray, num_classes: int) -> np.ndarray:
+    """ Calculate and return confusion matrix for the predicted and expected labels
```
blazekadam: You usually start on the next line with multi-line docstrings.

cxflow/utils/confusion_matrix.py
```diff
@@ -0,0 +1,20 @@
...
+
+    :param expected: array of expected classes (integers) with shape `[num_of_data]`
+    :param predicted: array of predicted classes (integers) with shape `[num_of_data]`
+    :param num_classes: number of classification classes
+    :return: confusion matrix (cm) with absolute values
+    """
+    assert np.issubclass_(expected.dtype.type, np.integer)
```
blazekadam: I would use `expected.dtype.kind == 'i'` but this works just as good. Not sure which one is better, just fyi.

cxflow/utils/confusion_matrix.py
```diff
@@ -0,0 +1,20 @@
...
+    assert np.issubclass_(predicted.dtype.type, np.integer)
+    assert expected.shape == predicted.shape
+    assert num_classes > np.max([predicted, expected])
+    assert np.min([predicted, expected]) >= 0
```
blazekadam: Some assertion message would be nice here (same applies for the assertions above). E.g.:
`assert np.min([predicted, expected]) >= 0, 'Confussion matrix can be calculated with non-negative inputs only'`

cxflow/hooks/save_cm.py
```diff
@@ -0,0 +1,152 @@
...
+
+from . import AccumulateVariables
+from ..types import EpochData
+from ..datasets import BaseDataset
+from ..utils import confusion_matrix
+
+
+class SaveConfusionMatrix(AccumulateVariables):
+    """
+    Compute and save/store confusion matrix figure for the predicted and expected labels.
```
blazekadam: I would add *after each epoch*.

cxflow/hooks/save_cm.py
```diff
@@ -0,0 +1,152 @@
...
+                classes_names = getattr(self._dataset, self._classes_names_method_name)()
+
+            # Get number of classes from possible sources
+            if classes_names:
+                num_classes = len(classes_names)
+            elif hasattr(self._dataset, self._num_classes_method_name):
+                num_classes = getattr(self._dataset, self._num_classes_method_name)()
+                assert num_classes > np.max([predicted, expected])
+            else:
+                num_classes = np.max([predicted, expected]) + 1
```
blazekadam: This is not tested :(

cxflow/hooks/save_cm.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def after_epoch(self, epoch_id: int, epoch_data: EpochData) -> None:
+        for stream_name, variables in self._accumulator.items():
+            predicted = np.array(variables[self._predictions_name])
+            expected = np.array(variables[self._labels_name])
+
+            # Try to get names of classes from possible sources
+            classes_names = False
+            if self._classes_names is not None:
+                classes_names = self._classes_names
```
blazekadam: This branch is not tested.

cxflow/tests/hooks/save_cm_test.py
bvavrik: It is utilized, because it is the default method for hook to get number of classes

PR #214, opened 2018-03-10
Add dataset stream_info method

blazekadam:


cxflow/datasets/base_dataset.py
```diff
@@ -45,3 +48,32 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
...
+                    row = [key, value_arr.dtype, value_arr.shape]
+
+                    if value_arr.dtype.kind in 'bui':  # boolean, unsigned, integer
+                        row.append('{} - {}'.format(value_arr.min(), value_arr.max()))
+                    elif value_arr.dtype.kind is 'f':
+                        row.append('{0:.2f} - {1:.2f}'.format(value_arr.min(), value_arr.max()))
+                    rows.append(row)
+                for line in tabulate.tabulate(rows, headers=['name', 'dtype', 'shape', 'range'],
+                                              tablefmt='grid').split('\n'):
+                    logging.info(line)
```
petrbel: Is logging really what we want? The information obtained by running this command is actually the desired output, not just some byproduct logging. I'd expect printing it to stdout which would allow users to do e.g. `cxflow dataset check config > check.txt` (without logging boilerplate).

cxflow/datasets/base_dataset.py
```diff
@@ -45,3 +48,32 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
...
+            except Exception:
+                logging.warning('Exception was raised during checking stream `%s`, '
+                                'add --verbose flag to see the stack trace', stream_name)
```
petrbel: hmm, the second part is a bit confusing when already running with `-v` flag

cxflow/datasets/base_dataset.py
```diff
@@ -45,3 +48,32 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
...
+                logging.debug(traceback.format_exc())
```
petrbel: Is this different from `logging.exception`? Not sure.

cxflow/datasets/base_dataset.py
```diff
@@ -45,3 +48,32 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
         :param kwargs: dataset configuration as ``**kwargs`` parsed from ``config['dataset']``
         :raise NotImplementedError: if not overridden
         """
+
+    def check(self) -> None:
```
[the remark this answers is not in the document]
blazekadam: Well, it actually checks
- if the streams can be created properly
- if the data is not ragged

I am not saying it is the best name for it but the alternatives are not that great as well. My thinking was that with this method, I am *checking* if the streams work properly; hence, the name.

Next, I plan on implementing `deep_check` which would iterate through the whole stream and `info` which would output some general info about the stream (its parents, public methods, and their docs).

So what alternative would you suggest?

cxflow/datasets/base_dataset.py
blazekadam: `logging.exception` would be displayed even with `INFO` level...

cxflow/datasets/base_dataset.py
blazekadam: I think that checking the logging level here is a bit overkill. How big is the `bit` exactly? You would see the exception right underneath this message...

cxflow/datasets/base_dataset.py
blazekadam: I really can not imagine a use case for piping it like you suggest. The output is not suitable for automated processing, some info is provided by logging so you would receive just a bunch of tables without context. I feel like you are overthinking it here. Its a simple dataset method intended for debugging and development, not an essential building block.

blazekadam:
Can you specify the changes you are requesting?

cxflow/datasets/base_dataset.py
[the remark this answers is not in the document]
FloopCZ: Here are a few alternatives to consider:
`stream_shape_info`
`print_stream_shape`
or just `stream_shape`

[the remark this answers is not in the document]
petrbel: or `streams`?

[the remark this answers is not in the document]
blazekadam: I would vote for `check` or `stream_info`. I kinda got attached to `check` though...

cxflow/datasets/base_dataset.py
```diff
@@ -45,3 +48,32 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
...
+                logging.info(stream_name)
+                batch = next(iter(stream_fn()))
+                rows = []
+                for key, value in batch.items():
+                    value_arr = np.array(value)
+                    if value_arr.ndim > 0 and value_arr.shape[1:] != np.array(value_arr[0]).shape:
+                        logging.warning('*** stream source `%s` appears to be ragged (non-rectangular) ***', key)
+                    row = [key, value_arr.dtype, value_arr.shape]
+
+                    if value_arr.dtype.kind in 'bui':  # boolean, unsigned, integer
```
FloopCZ: Nice comment. :+1:

cxflow/datasets/base_dataset.py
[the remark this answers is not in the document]
FloopCZ: From these two, I like `stream_info` more. `check` is too generic. What about @petrbel ?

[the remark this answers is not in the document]
petrbel: I agree with @FloopCZ

blazekadam:
name updated to `stream_info`

FloopCZ:
> it is not AbstractDataset but BaseDataset

The naming was not the point, it was about adding functionality to parent classes.

> the user is not forced to inherit from it unless he or she wants this feature

But he is forced to inherit from it if he or she wants this feature. :-) Even though it uses no information from the object itself. It is similar to `numpy` - if you want to utilize `np.sum`, you don't have to inherit from `ndarray`, you just need to be `iterable`. I like concept checking more than inheritance and prefer non-intrusive approach. But is is just an opinion.

> having cxflow as a dependency when implementing a cxflow dataset is not an issue imo

Is not a _big_ problem. It is just kind of inconvenient when you implement native datasets. But you can always wrap them and use some metaclass magic, right...

> multi-inheritance is not an anti-pattern, or is it?

Tell that to Java programmers! :laughing: 

> integrating this feature to cxflow cli is much more complicated, especially when it is not tested and about to be reworked

I know, it is more complicated. I just wanted to say that I like it better.

blazekadam:
Well, I can see your points (especially the `numpy` one); nonetheless, I think we may settle with this option (considering we already extend it with our private `CXDataset`). Let's postpone refactoring this feature to CLI with the CLI refactor.

cxflow/datasets/base_dataset.py
```diff
@@ -52,7 +52,8 @@ def _configure_dataset(self, output_dir: Optional[str], **kwargs):
 
     def stream_info(self) -> None:
         """Check and report source names, dtypes and shapes of all the streams available."""
-        stream_names = [stream_name for stream_name in dir(self) if 'stream' in stream_name]
+        stream_names = [stream_name for stream_name in dir(self)
+                        if 'stream' in stream_name and stream_name != 'stream_info']
```
petrbel: haha :D

blazekadam: nothing funny here :)

blazekadam:
@petrbel are you OK with the changes?

petrbel:
@blazekadam the tests seem to fail, though

blazekadam:
@petrbel yea well that has to be obviously fixed, nonetheless, my question holds...

PR #221, opened 2018-03-27
Add cxflow eval command

### LINKED ISSUE #76: Allow to predict other streams or add `cxflow eval`
Cxflow predict is meant to evaluate raw data without the ground truth or training data.
However, currently, cxflow predict calls extra streams. This is unfortunate from multiple reasons:

1. It is the only way to evaluate valid and test streams. However, I would like to be able to evaluate valid and test streams without loading new data or implementing prediction stream.
2. I don't need to own the valid or training data in a production ready environment where predict is meant to be used. I know that valid and test streams can be skipped in the config, but still, it is not intuitive that `predict` calls `valid_stream` and `test_stream`.

Solutions:
The most obvious solution to both issues is to have two functions, e.g., `cxflow test` and `cxflow predict`. I am not sure about the name for `cxflow test`, since it runs `extra_streams`. `cxflow extra` does not sound proper either. Maybe we could rename `extra_streams` to `test_streams` or similar.

### LINKED ISSUE #213: Name one of the resume arguments
Using `cxflow resume` with additional arguments is complicated as one has to specify the log dir twice.
```
cxflow resume log/my_dir log/my_dir my_argument=12
```

This is because both the config dir and checkpoint dir are positional arguments. We should name one of them and make it optional.

### LINKED ISSUE #220: Allow prediction of any stream
Following up todays discussion with @blazekadam.

Prediction CLI syntax should look as follows:
`cxflow predict <stream>`
and the predict section in the config file should look as follows:
```
predict:
  valid:
    dataset: ...
    model: ...
    hooks: ...
  production:
    dataset: ...
    model: ...
    hooks: ...
```

The reasons are obvious.

blazekadam:
Add `cxflow eval <stream_name> ...` command.

- allow to eval arbitrary stream (with config updates in `eval.<stream_name>` config section)
- keep support for `cxflow predict ...` command and `predict` config section (with deprecation warnings)
- name `--config, -c` argument and make it optional (only the `model_path` has to be specified)

In other words (commands):
`cxflow predict log/dir` becomes `cxflow eval predict log/dir`
`cxflow predict dir/config log/model` becomes `cxflow eval predict log/model --config dir/config`
`cxflow predict production log/model` is now available

*tests are omitted as CLI is about to be reworked (I tested it manually though)*

closes #220 
closes #213 
closes #76 (already closed)


cxflow/cli/eval.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    config = None
+
+    try:
+        model_dir = path.dirname(model_path) if not path.isdir(model_path) else model_path
+        config_path = find_config(model_dir if config_path is None else config_path)
+        config = load_config(config_file=config_path, additional_args=cl_arguments)
+
+        if stream_name == CXF_PREDICT_STREAM and stream_name in config:  # old style ``cxflow predict ...``
+            logging.warning('Old style ``predict`` configuration section is deprecated and will not be supported, '
```
FloopCZ: `,` -> `.`

cxflow/cli/eval.py
```diff
@@ -0,0 +1,100 @@
...
+                            'Use ``eval.predict`` section instead.')
+            config['eval'] = {'predict': config['predict']}
+
+        if 'eval' in config and stream_name in config['eval']:
+            update_section = config['eval'][stream_name]
+            for subsection in ['dataset', 'model', 'main_loop']:
+                if subsection in update_section:
+                    for key in update_section[subsection]:
+                        config[subsection][key] = update_section[subsection][key]
```
FloopCZ: Couldn't we just
```
config[subsection].update(update_section[subsection])
```
?

cxflow/cli/eval.py
```diff
@@ -0,0 +1,100 @@
...
+    try:
+        config_path = find_config(config_path)
+        restore_from = restore_from or path.dirname(config_path)
+        config = load_config(config_file=config_path, additional_args=cl_arguments)
+
+        if 'predict' in config:
+            for section in ['dataset', 'model', 'main_loop']:
+                if section in config['predict']:
+                    for key in config['predict'][section]:
+                        config[section][key] = config['predict'][section][key]
```
FloopCZ: ditto

cxflow/cli/eval.py
```diff
@@ -0,0 +1,100 @@
...
+            if 'hooks' in config['predict']:
+                config['hooks'] = config['predict']['hooks']
+            else:
+                logging.warning('Config does not contain `predict.hooks` section. No hook will be employed during the prediction.')
```
FloopCZ: line too long

cxflow/cli/eval.py
```diff
@@ -0,0 +1,100 @@
...
+    :param output_root: output root in which the training directory will be created
+    """
+    config = None
+
+    try:
+        model_dir = path.dirname(model_path) if not path.isdir(model_path) else model_path
+        config_path = find_config(model_dir if config_path is None else config_path)
+        config = load_config(config_file=config_path, additional_args=cl_arguments)
+
+        if stream_name == CXF_PREDICT_STREAM and stream_name in config:  # old style ``cxflow predict ...``
```
FloopCZ: I believe this condition will never be satisfied, sinc `entry_point.py:L64` calls predict, not evaluate. I suggest to remove this condition and only allow the new style call in this function.

cxflow/main_loop.py
```diff
@@ -305,10 +305,14 @@ def training():
 
         self._try_run(training)
 
-    def run_prediction(self) -> None:
-        """Run the main loop for in the prediction mode."""
+    def run_evaluation(self, stream_name: str=CXF_PREDICT_STREAM) -> None:
```
FloopCZ: I suggest to make the `stream_name` argument obligatory. We always pass the argument anyway and soon, we will drop the constant whatsoever.

cxflow/cli/eval.py
blazekadam: will be removed

PR #222, opened 2018-03-28
Add temporary fix for YAML ordereddict

FloopCZ:
Sometimes when expanding ruamel's `ordereddict` via `**` or directly converting it to `dict`, some values just get lost. This PR is a temporary fix, but the issue should be reported upstream.

Minimal working example to reproduce `ruamel`'s bug:
```python
#!/usr/bin/env python3
import ruamel.yaml

code = """\
anchored: &anchor
  a : 1

mapping:
  <<: *anchor
  b: 2
"""

mapping = ruamel.yaml.load(code, ruamel.yaml.RoundTripLoader)['mapping']

print(mapping)
print({**mapping})
print(dict(mapping))
print(dict(mapping.items()))
```

Output:
```
ordereddict([('b', 2), ('a', 1)])
{'b': 2}
{'b': 2}
{'b': 2, 'a': 1}
```

### DIFF
--- a/cxflow/cli/common.py
+++ b/cxflow/cli/common.py
@@ -181,6 +181,7 @@ def create_hooks(config: dict, model: AbstractModel,
             if hook_params is None:
                 logging.warning('\t\t Empty config of `%s` hook', hook_path)
                 hook_params = {}
+            hook_params = dict(hook_params.items())
 
             hook_module, hook_class = parse_fully_qualified_name(hook_path)
             # find the hook module if not specified
--- a/cxflow/cli/common.py
+++ b/cxflow/cli/common.py
@@ -110,6 +110,10 @@ def create_model(config: dict, output_dir: Optional[str], dataset: AbstractDatas
     logging.info('Creating a model')
 
     model_config = config['model']
+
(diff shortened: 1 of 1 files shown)


FloopCZ:
Track https://bitbucket.org/ruamel/yaml/issues/196/loosing-items-on-cast-to-dict for more info.

petrbel:
Is this still WIP? I think we can merge it as it will take time untill ruamelyaml is updated.

FloopCZ:
Yes, there are multiple places in our code, where this should be fixed. And there should also be some comment documenting why do we do such an operation.

blazekadam:
Good job finding this out and creating the minimal working example. Lets wait a few days and see how the package maintainers will cope with it. I would rather wait a few weeks than hotfixing it in our code...

FloopCZ:
I would also rather wait. Unfortunately, even if this gets fixed upstream, it will take like two years before the fix propagates into commonly used non-rolling linux distributions. :-( We should hotfix our code anyway.

blazekadam:
@FloopCZ thank you for adding the comments

I would merge it since there is no response in the ruamel issue tracker yet.

PR #225, opened 2018-04-04
Add test for dataset stream_info method

gdynusa:
test logging of source names, dtypes and shapes of all the streams available in given dataset


gdynusa:
Created by mistake

PR #230, opened 2018-05-05
Add test for download utility

gdynusa:
Test downloading of datasets from given URLs and its possible unpacking

- [ ] Provide more accurate error/logging if some URL does not exist (@blazekadam)

- [ ] Host the data files (@blazekadam)


cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Define setUp as class so it is called only once and not before each method.
+        Define data_root and url variables.
+        """
+        super(DownloadTest, cls).setUpClass()
+
+        cls.data_root = '/root/cxflow/cxflow/tests/utils/download'
```
blazekadam: Not that this does not work but we usually define test constants like this without `setUp` and `tearDown` methods. I.e.:
```
class MyTestCase:
    _DATA_ROOT = '/root/...'
    _URL_ZIP = 'dfdsfdsf'
```

or even module level constants (if we want to import them in other test case) like

```
_DATA_ROOT = '/root/...'
_URL_ZIP = 'dfdsfdsf'

class MyTestCase:
    pass
```

which is a bit more lucid (and not `unittest` specific).

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+        cls.url_zip = 'http://gstore.unm.edu/apps/rgis/datasets/' \
```
blazekadam: Not sure if we want to rely on files hosted by third parties, we can host some files ourselves or use github hosted files like this one `https://github.com/Cognexa/cxflow/archive/v0.12.0.zip`. Regarding `zip` and `none` options, I ll find some suitable files...

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        """
+        Define tearDown as class so it is called only once and not after each method.
+        Delete created folders and files.
+        """
+        super(DownloadTest, cls).tearDownClass()
+
+        shutil.rmtree(cls.data_root)
```
blazekadam: What about our https://github.com/Cognexa/cxflow/blob/188bc74e6e9466c12f7595d90813c3cf6dc3ee67/cxflow/tests/test_core.py#L30

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        with LogCapture(level=logging.INFO) as log_capture:
+            download.maybe_download_and_extract(self.data_root, self.url_zip)
+
+        log_capture.check(
+            ('root', 'INFO', '\tdownloading ' + self.data_root + self.url_zip_base)
+            )
+
+        self.assertTrue(os.path.exists(self.data_root + self.url_zip_base))
+        self.assertTrue(os.path.exists(self.data_root + '/nps_boundary.shp'))
```
blazekadam: use `os.path.join` for joining file paths (it is platform independent)

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Define setUp as class so it is called only once and not before each method.
+        Define data_root and url variables.
+        """
+        super(DownloadTest, cls).setUpClass()
+
+        cls.data_root = '/root/cxflow/cxflow/tests/utils/download'
+        cls.url_zip = 'http://gstore.unm.edu/apps/rgis/datasets/' \
+            '7bbe8af5-029b-4adf-b06c-134f0dd57226/nps_boundary.original.zip'
+        cls.url_zip_base = '/' + os.path.basename(cls.url_zip)
```
blazekadam: always use `os.path.join` for joining file paths (it is platform independent)

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Define tearDown as class so it is called only once and not after each method.
+        Delete created folders and files.
+        """
+        super(DownloadTest, cls).tearDownClass()
+
+        shutil.rmtree(cls.data_root)
+
+    def test_download_and_unpack_successful(self):
+        """
```
blazekadam: You may shorten one-liner docstrings to (it is up to your taste though)
```
"""Your docstring."""
```

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+            download.maybe_download_and_extract(self.data_root, self.url_zip)
+
+        log_capture.check(
+            ('root', 'INFO', '\tdownloading ' + self.data_root + self.url_zip_base)
+            )
+
+        self.assertTrue(os.path.exists(self.data_root + self.url_zip_base))
+        self.assertTrue(os.path.exists(self.data_root + '/nps_boundary.shp'))
+
+    def test_download_fails(self):
```
blazekadam: the name is kinda misleading, the download actually succeeds, it is more about `extraction_fail`, is it not?

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+        log_capture.check(
+            ('root', 'INFO', '\tdownloading ' + self.data_root + self.url_rar_base),
+            ('root', 'INFO', 'File `' + self.data_root + self.url_rar_base + '`'
+                ' could not be extracted by `shutil.unpack_archive`. Please process it manually.')
+            )
+        
+        self.assertTrue(os.path.exists(self.data_root + self.url_rar_base))
+        self.assertFalse(os.path.exists(self.data_root + '/anomalousTrafficTest.txt'))
+        self.assertRaises(shutil.ReadError)
```
blazekadam: not sure what this line should do?

cxflow/tests/utils/download_test.py
blazekadam: or the download indeed fails?

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+        with LogCapture(level=logging.INFO) as log_capture:
+            download.maybe_download_and_extract(self.data_root, self.url_rar)
+
+        log_capture.check(
+            ('root', 'INFO', '\t`' + self.data_root + self.url_rar_base + '` already exists; skipping')
+            )
+
+    def test_url_doesnt_contain_filetype(self):
+        """
+        Given URL does not contain filetype, therefore does not download correctly and cannot be unpacked.
```
blazekadam: again misleading docstring, I suppose the file will be downloaded but wount be extracted?

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,121 @@
...
+            ('root', 'INFO', '\tdownloading ' + self.data_root + self.url_none_base),
+            ('root', 'INFO', 'File `' + self.data_root + self.url_none_base + '`'
+                ' could not be extracted by `shutil.unpack_archive`. Please process it manually.')
+            )
+
+        self.assertRaises(ValueError)
+
+    def test_url_ends_with_slash(self):
+        """
+        Given URL does end with slash, therefore self.data_root matches filepath and nothing is downloded.
```
blazekadam: This is one interesting edge-case. I will need to think about what is the expected behavior here...

cxflow/tests/utils/download_test.py
gdynusa: The download indeed fails, as given URL is invalid. Maybe we should raise exception or log something to let user know the problem is not extraction but failed download (which could happen for various reasons) ?

blazekadam: I see, in that case, the `logging` provided by the function is sort of misleading. I ll fix it.

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,118 @@
...
+            ('root', 'INFO', '\tdownloading ' + os.path.join(_DATA_ROOT, _URL_NONE_BASE)),
+            ('root', 'INFO', 'File `' + os.path.join(_DATA_ROOT, _URL_NONE_BASE) + '`'
+                ' could not be extracted by `shutil.unpack_archive`. Please process it manually.')
+            )
+
+        self.assertRaises(ValueError)
+
+    def test_url_ends_with_slash(self):
+        """
+        Given URL ends with slash, therefore _DATA_ROOT matches filepath and nothing is downloded.
```
FloopCZ: downloaded

cxflow/tests/utils/download_test.py
```diff
@@ -0,0 +1,118 @@
+import logging
+import os
+import shutil
+from testfixtures import LogCapture
+from cxflow.tests.test_core import CXTestCaseWithDir
+
+import cxflow.utils.download as download
+
+
+_DATA_ROOT = '/root/cxflow/cxflow/tests/utils/download'
```
FloopCZ: Users don't have the permission to write to this folder. Why do you change the TMP dir?

blazekadam: Furthermore, the test just became platform specific (with hard-coded slashes `/`).

I suppose there is a slight misunderstanding on how the `CXTestCaseWithDir` works. For each test, it prepares a clean directory which is saved in `self.tmpdir`. There is no need to override it or create your own.

@FloopCZ the tests in CI actually run under root :o, that is why it works I suppose

gdynusa: To be honest, I'm not sure why I left the directory specified, I guess it was a combination of two things, being relict of previous code together with rather irrational 'need' to know the location where folder gets greated :) Repaired.

blazekadam:
Close in favor of #233

PR #233, opened 2018-07-03
Updated download test

gdynusa:
closes #230


cxflow/utils/download.py
```diff
@@ -17,7 +17,11 @@ def maybe_download_and_extract(data_root: str, url: str) -> None:
     # make sure data_root exists
     os.makedirs(data_root, exist_ok=True)
 
-    filename = os.path.basename(url)
+    # create filename from url
+    if url.split('/')[-1] != '':
```
blazekadam: Although it is not likely, this can still fail... Imagine the following url `https://domain.com/get_file.php?file=asd94//`. Both `-1` and `-2` items would be empty strings after the split. 

Another funny thing is that the url can contain characters which can not be present in the filename, E.g.: `?` in the aforementioned example if you would go further to `-3`. 

What about
```python
for part in reversed(url.split('/')):
   filename = sanitize(part)
   if len(filename) > 0:
      break
else:
   raise ValueError('Some messsage')
```

wherein the sanitize function can be str replace of non alphanum characters or something.

cxflow/tests/utils/download_test.py
```diff
@@ -55,6 +57,20 @@ def test_download_fails(url, url_base, tmpdir, caplog):
     assert not os.path.exists(os.path.join(tmpdir, url_base))
 
 
+def test_url_sanitized(tmpdir, caplog):
+    """Only valid characters used to create filename."""
+    caplog.set_level(logging.INFO)
+    download.maybe_download_and_extract(tmpdir, _URL_TO_SANITIZE)
+
+    assert 'n00._b' in caplog.text
```
blazekadam: This is a rather vague assertion. The given string can be logged in various context but whatever. Just be aware of it.

cxflow/tests/utils/download_test.py
```diff
@@ -12,6 +12,8 @@
 _URL_NONE = 'https://github.com/Cognexa/cxflow-examples/releases/download/example-files/worldwide-companies-r10k'
 _URL_NONE_BASE = os.path.basename(_URL_NONE)
 _URL_INVALID = 'not.valid'
+_URL_TO_SANITIZE = 'https://n%0 ?0.=_b//'
+_URL_RAISE_ERROR = '://*%~/'
```
blazekadam: what an url :)

cxflow/utils/download.py
```diff
@@ -17,11 +18,13 @@ def maybe_download_and_extract(data_root: str, url: str) -> None:
     # make sure data_root exists
     os.makedirs(data_root, exist_ok=True)
 
-    # create filename from url
-    if url.split('/')[-1] != '':
-        filename = url.split('/')[-1]
+    # create sanitized filename from url
+    for part in reversed(url.split('/')):
+        filename = re.sub(r'[^a-zA-Z0-9_.\-]', '', part)
```
blazekadam: image

cxflow/tests/utils/download_test.py
gdynusa: I'm only validating filename was sanitized correctly here. I guess I could also assert as in previous test case that nothing was downloaded.

blazekadam: I know, it is just kinda of indirect.

What you want to test is:
- for the given ugly url, is the filename well sanitized?

What you actually test is:
- for the given url, does the log contain some part that looks like the filename I expect?

Do you see the difference? You may log the sanitized name but use some other for the actual filename etc. etc. etc.

The best course would be adding a simple separated function `sanitize` which would take an url and return the filename or raise an error. You would be able to properly test it and code in `maybe_download...` would be cleaner.

blazekadam: Btw having difficulties to unit test something properly may be a good sign of design flaws (sub-optimal structure, function args and return values, function side-effects etc.).

cxflow/utils/download.py
```diff
@@ -7,6 +7,24 @@
 import re
 
 
+def sanitize(url: str) -> str:
```
blazekadam: `sanitize_url` would be a better name imo (do not underestimate the importance of proper, accurate and consistent naming).

cxflow/utils/download.py
```diff
@@ -7,6 +7,24 @@
...
+    """
+    Sanitize url so it can be used to create valid filename.
+
+    :param url: url to create filename from
+    :return: created filename
```
blazekadam: You can add `:raise ValueError: when the given url can not be sanitized` line before the `:return: ...` directive.

cxflow/tests/utils/download_test.py
```diff
@@ -62,7 +62,7 @@ def test_url_sanitized(tmpdir, caplog):
     caplog.set_level(logging.INFO)
     download.maybe_download_and_extract(tmpdir, _URL_TO_SANITIZE)
 
-    assert 'n00._b' in caplog.text
+    assert 'downloading ' + os.path.join(tmpdir, 'n00._b') in caplog.text
```
blazekadam: Apart from this line, please test the new function directly.

cxflow/utils/download.py
```diff
@@ -7,6 +7,24 @@
 import re
 
 
+def sanitize(url: str) -> str:
+    """
+    Sanitize url so it can be used to create valid filename.
```
blazekadam: I am no native speaker but 

*Sanitize the given url so that it can be used as a filename.*

reads a bit better

cxflow/utils/download.py
```diff
@@ -19,12 +37,7 @@ def maybe_download_and_extract(data_root: str, url: str) -> None:
     os.makedirs(data_root, exist_ok=True)
 
     # create sanitized filename from url
-    for part in reversed(url.split('/')):
-        filename = re.sub(r'[^a-zA-Z0-9_.\-]', '', part)
-        if len(filename) > 0:
-            break
-    else:
-        raise ValueError('Could not create reasonable name for file from url %s', url)
+    filename = sanitize(url)
```
blazekadam: This looks much better! Good job.

petrbel:
Wow, you guys finally merged this! :D

gdynusa:
I so wanna react with @FloopCZ head icon ;(

PR #235, opened 2018-07-25
Cati hooks

gdynusa:


cxflow/hooks/flatten.py
```diff
@@ -0,0 +1,47 @@
...
+                        if None, the hook will be applied to all the available streams
+        """
+        super().__init__(**kwargs)
+        self._variables = variables
+        self._streams = streams
+
+    def after_batch(self, stream_name: str, batch_data: Batch) -> None:
+        if self._streams is None or stream_name in self._streams:
+            for variable in self._variables:
+                assert variable in batch_data
```
petrbel: I suggest adding some message explaining what is wrong
```
assert <condition>, "message"
```

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+    is named by the corresponding stream source.
+
+    .. code-block:: yaml
+        :caption: Save a csv with columns `red`, `green`, and `blue` to `/var/colors.csv`.
+                  The stream variable `color` is expected to be a sequence of three numbers.
+
+        hooks:
+          - LogitsToCsv:
+              variable: color
+              class_names: ['red', 'green', 'blue']
```
petrbel: Are string `'` required in YAML? The other entries don'thave them...

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+        self._variable = variable
+        self._class_names = class_names
+        self._id_variable = id_variable
+        self._output_file = output_file
+        self._streams = streams
+        self._accumulator = []
+
+    def after_batch(self, stream_name: str, batch_data: Batch) -> None:
+        """
+        Accumulate the given logits.
```
petrbel: such a short docstring can/should be oneliner

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        if self._streams is not None and stream_name not in self._streams:
+            return
+
+        # Assert variables in batch data.
+        assert self._id_variable in batch_data
```
petrbel: again, messages would be useful

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+        assert self._variable in batch_data
```
petrbel: again

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Assert equal batch sizes.
+        assert len(batch_data[self._id_variable]) == len(batch_data[self._variable])
```
petrbel: again

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Iterate through examples.
+        for example_idx, example_id in enumerate(batch_data[self._id_variable]):
+            assert len(batch_data[self._variable][example_idx]) == len(self._class_names)
```
petrbel: again

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+            # Build the csv record.
+            record = OrderedDict()
+            record[self._id_variable] = example_id
+            for class_idx, class_name in enumerate(self._class_names):
+                record[class_name] = batch_data[self._variable][example_idx][class_idx]
+            self._accumulator.append(record)
+
+    def after_epoch(self, epoch_id: int, **_) -> None:
+        """
+        Save all the accumulated data to csv.
```
petrbel: dtto

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        if self._accumulator != []:
```
petrbel: What about `len(acc) > 0`? In case the accumulator implemntation changes (which can surely happen), this could will not be valid.

cxflow/hooks/logits_to_csv.py
```diff
@@ -0,0 +1,84 @@
...
+            logging.info('Saving logits from '.join(self._variable) + ' to %s.', self._output_file)
```
petrbel: Can `self._variable` be an iterable? Why is `join` used?

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,150 @@
...
+                  batches (from the train stream).
+
+        hooks:
+          - PlotLines:
+              variables: [xs, ys]
+              example_count: 2
+              batch_count: 10
+    """
+
+    _ROOT_DIR = 'visual'
```
petrbel: this should definitely be configurable via constructor

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,150 @@
...
+        return fig
+
+    def after_batch(self, stream_name: str, batch_data: Batch):
+        """
+        Save images in provided streams from selected variable. The amount of batches and images to be processed is
+        possible to control by ``batch_count`` and ``example_count`` parameters.
+        """
+        if self._streams is None or stream_name in self._streams:
+            # assert variables in batch data
+            assert self._id_variable in batch_data
```
petrbel: I still like messages

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,150 @@
...
+            assert self._pad_mask_variable is None or self._pad_mask_variable in batch_data
+            for variable in self._variables:
+                assert variable in batch_data
```
petrbel: dttop

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def after_batch(self, stream_name: str, batch_data: Batch):
+        """
+        Save images in provided streams from selected variable. The amount of batches and images to be processed is
+        possible to control by ``batch_count`` and ``example_count`` parameters.
+        """
+        if self._streams is None or stream_name in self._streams:
+            # assert variables in batch data
+            assert self._id_variable in batch_data
+            assert self._pad_mask_variable is None or self._pad_mask_variable in batch_data
```
petrbel: dtto

cxflow/hooks/sequence_to_csv.py
```diff
@@ -0,0 +1,106 @@
...
+        self._accumulator = []
+
+    def after_batch(self, stream_name: str, batch_data: Batch) -> None:
+        """
+        Accumulate the given sequences.
+        """
+        if self._streams is not None and stream_name not in self._streams:
+            return
+
+        assert len(self._variables) > 0
```
petrbel: message

cxflow/hooks/sequence_to_csv.py
```diff
@@ -0,0 +1,106 @@
...
+        self._pad_mask_variable = pad_mask_variable
+        self._streams = streams
+        self._accumulator = []
+
+    def after_batch(self, stream_name: str, batch_data: Batch) -> None:
+        """
+        Accumulate the given sequences.
+        """
+        if self._streams is not None and stream_name not in self._streams:
+            return
```
petrbel: I like this condition very much

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,150 @@
...
+        ax.legend()
+        fig.tight_layout()
+        return fig
+
+    def after_batch(self, stream_name: str, batch_data: Batch):
+        """
+        Save images in provided streams from selected variable. The amount of batches and images to be processed is
+        possible to control by ``batch_count`` and ``example_count`` parameters.
+        """
+        if self._streams is None or stream_name in self._streams:
```
petrbel: consider using the negation of this condition containing only `return`. hen, this code wouldn't be indented which I like better.

cxflow/hooks/flatten.py
```diff
@@ -0,0 +1,47 @@
...
+        :param variables: names of the variables to be flattened
+        :param streams: list of stream names to be considered;
+                        if None, the hook will be applied to all the available streams
+        """
+        super().__init__(**kwargs)
+        self._variables = variables
+        self._streams = streams
+
+    def after_batch(self, stream_name: str, batch_data: Batch) -> None:
+        if self._streams is None or stream_name in self._streams:
```
petrbel: consider using the negation of this condition containing only `return`. hen, this code wouldn't be indented which I like better.

cxflow/hooks/sequence_to_csv.py
```diff
@@ -0,0 +1,106 @@
...
+                record['index'] = seq_idx
+                for variable in self._variables:
+                    record[variable] = batch_data[variable][example_idx][seq_idx]
+                self._accumulator.append(record)
+
+    def after_epoch(self, epoch_id: int, **_) -> None:
+        """
+        Save all the accumulated data to csv.
+        """
+        if self._accumulator != []:
```
petrbel: dtto

cxflow/hooks/sequence_to_csv.py
```diff
@@ -0,0 +1,106 @@
...
+            logging.info('Saving ' + ', '.join(self._variables) + ' to %s.', self._output_file)
```
petrbel: dtto

requirements.txt
```diff
@@ -7,4 +7,9 @@ typing>=3.5.3.0
...
 tabulate>=0.8.1
 babel>=2.5.1
-matplotlib>=2.1.1
\ No newline at end of file
+matplotlib>=2.1.1
+pandas
+
+# testing
+pytest
+
```
petrbel: useless line

requirements.txt
```diff
@@ -7,4 +7,9 @@ typing>=3.5.3.0
 testfixtures>=5.1.1
 tabulate>=0.8.1
 babel>=2.5.1
-matplotlib>=2.1.1
\ No newline at end of file
+matplotlib>=2.1.1
+pandas
```
petrbel: missing version

FloopCZ:
@petrbel Thank you for the nice review (of my code :-D ). I basically agree with everything, I am not so strict about the assertion messages as long as the assertion is simple, but better messages would definitely not hurt. Also, assertions should not be used like this at all, we should instead raise `ValueError` like we do in the rest of `cxflow`. Assertions are only for internal checks, but when something can happen because of an invalid user input, we should `raise` instead.

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,160 @@
...
+        hooks:
+          - PlotLines:
+              variables: [xs, ys]
+              example_count: 2
+              batch_count: 10
+    """
+
+    def __init__(self, output_dir: str, variables: Iterable[str], streams: Optional[Iterable[str]]=None,
+                 id_variable: str='ids', pad_mask_variable: Optional[str]=None, out_format: str='png',
+                 ymin: float=None, ymax: float=None, example_count: int=None, batch_count: int=None,
```
petrbel: `ymin`, `ymax`, `example_count` and `batch_count` should be of type `Optional[...]`

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # create the output directory
+        stream_out_dir = os.path.join(self._output_dir, self._root_dir,
+                                      'epoch_{}'.format(self._current_epoch_id), stream_name)
+        os.makedirs(stream_out_dir, exist_ok=True)
+
+        # iterate through the examples and generate plots
+        for i, ex_id in enumerate(batch_data[self._id_variable]):
+            # only plot the requested number of examples
+            if self._example_count and i + 1 > self._example_count:
```
petrbel: why not `i >= self._example_count` instead? I think it's more intuitive

cxflow/tests/hooks/flatten_test.py
```diff
@@ -0,0 +1,62 @@
...
+def test_flattening_variables():
+    """Test flattening selected variables when hook applied to all available streams."""
+
+    selected_vars = OrderedDict([('1d', '1d_flat'), ('2d', '2d_flat'), ('3d', '3d_flat')])
+    expected_flat_vars = [np.array([1.11]), np.array(range(10)), np.array(range(20))]
+    flatten_vars = Flatten(variables=selected_vars)
+
+    for _ in range(_ITERS):
+        batch = get_batch()
+        flatten_vars.after_batch(_STREAM_NAME, batch)
```
petrbel: we should also call `after_batch` to another stream and check that the changes are applied only to `_STREAM_NAME`

cxflow/tests/hooks/flatten_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+    for var_flat, exp_flat in zip(selected_vars.values(), expected_flat_vars):
+        assert np.array_equal(batch[var_flat], exp_flat)
+
+
+def test_flattening_variables_stream_not_in_specified():
+    """Test flattening selected variables is not done when stream is not in available streams."""
+
+    selected_vars = OrderedDict([('1d', '1d_flat'), ('3d', '3d_flat')])
```
petrbel: in this case you omit `2d`, so it should be tested that `2d_flat` is not present in the output

cxflow/tests/hooks/logits_to_csv_test.py
```diff
@@ -0,0 +1,84 @@
...
+    logits_to_csv = LogitsToCsv(variable=selected_var, class_names=_CLASSES, id_variable=id_var, output_file=filename,
+                                streams=['test'])
+
+    for i in range(_ITERS):
+        batch = get_batch()
+        logits_to_csv.after_batch(_STREAM_NAME, batch)
+
+    logits_to_csv.after_epoch(0)
+
+    assert logits_to_csv._accumulator == []
```
petrbel: again, I'd prefer asserting that the length is zero since the acc. implementation may change in the future to another iterable

cxflow/tests/hooks/plot_lines_test.py
```diff
@@ -0,0 +1,130 @@
...
+    suffix = '-vs-'.join(selected_vars)
+    plot_lines = PlotLines(output_dir=tmpdir, variables=selected_vars)
+
+    for i in range(_ITERS):
+        batch = get_batch()
+        plot_lines.after_batch(_STREAM_NAME, batch)
+
+        for id_var in batch['ids']:
+            filename = '{}_batch_{}_plot-{}.{}'.format(id_var, i+1, suffix, 'png')
+            assert os.path.exists(os.path.join(tmpdir, 'visual', 'epoch_{}'.format(_EPOCH_ID), _STREAM_NAME, filename))
```
petrbel: Just to be sure: you're also checking the default value of `root_dir`, right? Is this intentional?

In this case I'd rather specify it in the constructor as well...

cxflow/tests/hooks/plot_lines_test.py
```diff
@@ -0,0 +1,130 @@
...
+                           batch_count=batches, root_dir=root_dir)
+
+    for _ in range(_ITERS):
+        batch = get_batch()
+        plot_lines.after_batch(_STREAM_NAME, batch)
+
+    for i in range(batches):
+        for e, id_var in zip(range(_EXAMPLES), batch[id_variable]):
+            filename = '{}_batch_{}_plot-{}.{}'.format(id_var, i+1, suffix, out_format)
+            if e < examples:
```
petrbel: this whole if/else could be replaced with a single line which would remove the duplicite condition of assert
```
assert (e < examples) == os.path.exists(os.path.join(tmpdir, root_dir, 'epoch_{}'.format(_EPOCH_ID), _STREAM_NAME, filename))
```

However, the readability is pretty low so I don't know what's better

cxflow/tests/hooks/plot_lines_test.py
```diff
@@ -0,0 +1,130 @@
...
+    for i_not in range(batches, _ITERS):
+        for id_not in batch['ids']:
+            filename = '{}_batch_{}_plot-{}.{}'.format(id_not, i_not+1, suffix, out_format)
+            assert not os.path.exists(
+                os.path.join(tmpdir, root_dir, 'epoch_{}'.format(_EPOCH_ID), _STREAM_NAME, filename))
+
+
+_VARS = [['not-there', 'cost'],
+         ['input', 'cost'],
+         ['input', 'cost']]
```
petrbel: you know you have a duplicite item, right?

cxflow/tests/hooks/flatten_test.py
gdynusa: @petrbel I'm not sure I understand. Isn't that something that can be tested for after_epoch hooks but not really for after_batch hooks as those get data straight from inside specified stream?

cxflow/tests/hooks/plot_lines_test.py
gdynusa: It's on purpose, the other two lists that are later zipped with this one differ, so it creates three unique inputs to parametrized test

cxflow/tests/hooks/plot_lines_test.py
gdynusa: That's a nice trick!

cxflow/tests/hooks/flatten_test.py
petrbel: tested elsewhere

cxflow/tests/hooks/flatten_test.py
```diff
@@ -34,11 +34,14 @@ def test_flattening_variables():
     for var_flat, exp_flat in zip(selected_vars.values(), expected_flat_vars):
         assert np.array_equal(batch[var_flat], exp_flat)
 
+    with pytest.raises(KeyError):
+        not_flattened = batch['2d_flat']
```
petrbel: `assert '2d_flat' not in batch`

cxflow/tests/hooks/plot_lines_test.py
```diff
@@ -50,7 +50,7 @@ def test_plotting_lines_with_optional_arguments(tmpdir):
 
     selected_vars = ['input', 'cost']
     id_variable = 'a_ids'
-    out_format = 'jpg'
```
petrbel: this is weird, @FloopCZ do you know anything about matplotlib dropping support for jpegs?

cxflow/tests/hooks/flatten_test.py
petrbel: `assert '2d_flat' not in batch`

cxflow/hooks/plot_lines.py
```diff
@@ -0,0 +1,160 @@
...
+        Hook constructor.
+
+        :param output_dir: output directory where plots will be saved
+        :param variables: names of the variables to be plotted
+        :param streams: list of stream names to be dumped; can be None to dump all streams
+        :param id_variable: name of the source which represents a unique example id
+        :param pad_mask_variable: name of the source which represents the padding mask
+        :param out_format: extension of the saved image
+        :param ymin: minimum on the Y axis
+        :param ymin: maximum on the Y axis
```
FloopCZ: ymin->ymax

